PR #3, opened 2024-01-28
Add GitHub actions

ArangoGutierrez:


.github/ISSUE_TEMPLATE
```diff
@@ -0,0 +1,37 @@
...
+ - [ ] The k8s-device-plugin container logs
+ - [ ] The kubelet logs on the node (e.g: `sudo journalctl -r -u kubelet`)
+
+Additional information that might help better understand your environment and reproduce the bug:
+ - [ ] Docker version from `docker version`
+ - [ ] Docker command, image and tag used
+ - [ ] Kernel version from `uname -a`
+ - [ ] Any relevant kernel output lines from `dmesg`
+ - [ ] NVIDIA packages version from `dpkg -l '*nvidia*'` _or_ `rpm -qa '*nvidia*'`
+ - [ ] NVIDIA container library version from `nvidia-container-cli -V`
```
elezar, line 37:
```suggestion
 - [ ] NVIDIA Container Toolkit version from `nvidia-ctk --version`
```

.github/ISSUE_TEMPLATE
```diff
@@ -0,0 +1,37 @@
...
+### 3. Information to [attach](https://help.github.com/articles/file-attachments-on-issues-and-pull-requests/) (optional if deemed irrelevant)
+
+Common error checking:
+ - [ ] The output of `nvidia-smi -a` on your host
+ - [ ] Your docker configuration file (e.g: `/etc/docker/daemon.json`)
+ - [ ] The k8s-device-plugin container logs
+ - [ ] The kubelet logs on the node (e.g: `sudo journalctl -r -u kubelet`)
+
+Additional information that might help better understand your environment and reproduce the bug:
+ - [ ] Docker version from `docker version`
```
elezar, line 32: Docker is probably no longer relevant for k8s components.

.github/workflows/images.yaml
```diff
@@ -0,0 +1,76 @@
...
+            GENERATE_ARTIFACTS="false"
+          elif [[ "${{ github.event_name }}" == "pull_request" && "${{ github.event.pull_request.head.repo.full_name }}" == "${{ github.repository }}" ]]; then
+            GENERATE_ARTIFACTS="true"
+          elif [[ "${{ github.event_name }}" == "push" ]]; then
+            GENERATE_ARTIFACTS="true"
+          fi
+          echo "PUSH_ON_BUILD=${GENERATE_ARTIFACTS}" >> $GITHUB_ENV
+          echo "BUILD_MULTI_ARCH_IMAGES=${GENERATE_ARTIFACTS}" >> $GITHUB_ENV
+
+      - name: Set up Docker Buildx
```
elezar, line 60: We need to set up Qemu here.

elezar (commented):
My one question is whether the addition of the issue template makes sense at the moment. Let's rather leave it out until we have a better idea of what we want from users. In the case of GPU operator operands this would be different from standalone components. We may want to direct users to the `gpu-operator` repo instead.

versions.mk
```diff
@@ -13,7 +13,12 @@
...
 VERSION ?= v0.6.5
+MODULE := github.com/NVIDIA/k8s-driver-manager
 
 vVERSION := v$(VERSION:v%=%)
 
 CUDA_VERSION := 12.3.0
+GOLANG_VERSION ?= 1.20.5
+
+BUILDIMAGE_TAG ?= devel-go$(GOLANG_VERSION)
+BUILDIMAGE ?=  ghcr.io/nvidia/k8s-test-infra:$(BUILDIMAGE_TAG)
```
elezar, line 24: Missing newline.

Makefile
```diff
@@ -0,0 +1,80 @@
...
+.shell:
+	$(DOCKER) run \
+		--rm \
+		-ti \
+		-e GOCACHE=/tmp/.cache/go \
+		-e GOMODCACHE=/tmp/.cache/gomod \
+		-v $(PWD):/work \
+		-w /work \
+		--user $$(id -u):$$(id -g) \
+		$(BUILDIMAGE)
```
elezar, line 80: Missing newline.

Makefile
```diff
@@ -0,0 +1,80 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+DOCKER   ?= docker
+MKDIR    ?= mkdir
+
+include $(CURDIR)/versions.mk
+
+ifeq ($(IMAGE_NAME),)
+REGISTRY ?= nvidia
```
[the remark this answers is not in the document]
ArangoGutierrez: took inspiration from https://github.com/NVIDIA/k8s-device-plugin/blob/main/Makefile#L22

.github/ISSUE_TEMPLATE
ArangoGutierrez: Removed the issue template file for now

go.mod
```diff
@@ -1,4 +1,4 @@
-module gitlab.com/nvidia/cloud-native/k8s-driver-manager
+module github.com/NVIDIA/k8s-driver-manager
```
elezar, line 1: Nit: We're changing the module name not location.

deployments/container/Dockerfile.ubi8
```diff
@@ -14,7 +14,7 @@
 
 ARG CUDA_VERSION
 ARG BASE_DIST
-ARG GOLANG_VERSION=1.20.2
+ARG GOLANG_VERSION=1.20.5
```
elezar, line 17: Let's use:
```suggestion
ARG GOLANG_VERSION=x.x.x
```
here so that we don't have to update it (and keep it in sync) when bumping the golang version.

versions.mk
```diff
@@ -17,3 +17,4 @@ VERSION ?= v0.6.5
 vVERSION := v$(VERSION:v%=%)
 
 CUDA_VERSION := 12.3.0
+GOLANG_VERSION ?= 1.20.5
```
elezar, line 21: There were some glibc issues when upgrading from `1.20.4` to `1.20.5` which may require that the dockerfiles themselves also be updated. Could we rather skip the golang bump in this PR and assess the impact separately?

.github/ISSUE_TEMPLATE
```diff
@@ -0,0 +1,37 @@
...
+### 2. Issue or feature description
+_Briefly explain the issue in terms of expected behavior and current behavior._
+
+### 3. Information to [attach](https://help.github.com/articles/file-attachments-on-issues-and-pull-requests/) (optional if deemed irrelevant)
+
+Common error checking:
+ - [ ] The output of `nvidia-smi -a` on your host
+ - [ ] Your docker configuration file (e.g: `/etc/docker/daemon.json`)
+ - [ ] The k8s-device-plugin container logs
+ - [ ] The kubelet logs on the node (e.g: `sudo journalctl -r -u kubelet`)
```
elezar: A lot of these are not applicable to this component. Let's rather remove this template until we have a better grasp of what we need.

.github/ISSUE_TEMPLATE
elezar, line 1: If this is removed in this commit ... let's rebase and remove the commit that adds this instead.

elezar (changes requested):

go.mod
ArangoGutierrez: so keep go.mod pointing to gitlab?

versions.mk
ArangoGutierrez: sure, yeah I remember those glibc issues

.github/ISSUE_TEMPLATE
ArangoGutierrez: will do

go.mod
ArangoGutierrez: I see, thanks

Makefile
```diff
@@ -0,0 +1,80 @@
...
+
+TARGETS := $(MAKE_TARGETS)
+
+DOCKER_TARGETS := $(patsubst %,docker-%, $(TARGETS))
+.PHONY: $(TARGETS) $(DOCKER_TARGETS)
+
+GOOS ?= linux
+
+build:
+	GOOS=$(GOOS) go build ./...
```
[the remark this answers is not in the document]
ArangoGutierrez: https://github.com/NVIDIA/k8s-driver-manager/issues/4

versions.mk
```diff
@@ -13,7 +13,12 @@
 # limitations under the License.
 
 VERSION ?= v0.6.5
+MODULE := github.com/NVIDIA/k8s-driver-manager
 
 vVERSION := v$(VERSION:v%=%)
 
 CUDA_VERSION := 12.3.0
+GOLANG_VERSION ?= 1.20.4
```
elezar, line 21: Any reason to not use `1.20.2` and update the go version in a follow-up?

ArangoGutierrez: not really, will fallback to .2

ArangoGutierrez: oh we might need a test-infra .2

ArangoGutierrez: https://github.com/NVIDIA/k8s-test-infra/pull/8

ArangoGutierrez: done

elezar (approved):
Thanks @ArangoGutierrez

PR #5, opened 2024-02-09
bump cuda base image and golang versions

tariq1890:


cdesiniotis (approved):

elezar (commented):
As a note 12.3.2 was just released.

We should really automate this using dependabot.

PR #8, opened 2024-02-28
add new CI step to validate go mod integrity

tariq1890:
Also includes minor yaml formatting improvements to the github actions file

### DIFF
--- a/.github/workflows/golang.yaml
+++ b/.github/workflows/golang.yaml
@@ -17,15 +17,27 @@ name: Golang
 on:
   pull_request:
     branches:
-      - master
-      - release-*
+    - master
+    - release-*
   push:
     branches:
-      - master
-      - release-*
+    - master
+    - release-*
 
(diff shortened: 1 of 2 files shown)


cdesiniotis (approved):

.github/workflows/golang.yaml
```diff
@@ -17,15 +17,27 @@ name: Golang
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Install Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: '1.20'
+      - run: make validate-modules
```
elezar, line 38: Just a note: I recently added logic to the device plugin to extract the go version from versions.mk so that these checks can be more consistent. Also added a containerized module validation. It would be good to iterate on something that can be used as standard across our projects.

PR #12, opened 2024-03-14
run go build in the cuda base image rather golang

tariq1890:
This change gets around the following error:

```
Auto upgrade policy of the GPU driver on the node cnt-server-2 is disabled
Cordoning node cnt-server-2...
node/cnt-server-2 cordoned
Draining node cnt-server-2 of any GPU pods...
nvdrain: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by nvdrain)
nvdrain: /lib64/libc.so.6: version `GLIBC_2.32' not found (required by nvdrain)
```

By running the `go build` in the runtime environment (cuda-base image), we remove the possibility of any execution failure from the glibc mismatch

### DIFF
--- a/deployments/container/Dockerfile.ubi8
+++ b/deployments/container/Dockerfile.ubi8
@@ -14,9 +14,8 @@
 
 ARG CUDA_VERSION
 ARG BASE_DIST
-ARG GOLANG_VERSION=x.x.x
 
-FROM golang:${GOLANG_VERSION} AS build
+FROM nvcr.io/nvidia/cuda:${CUDA_VERSION}-base-${BASE_DIST} AS build
 
 WORKDIR /work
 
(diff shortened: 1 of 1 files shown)


deployments/container/Dockerfile.ubi8
```diff
@@ -14,9 +14,8 @@
 
 ARG CUDA_VERSION
 ARG BASE_DIST
-ARG GOLANG_VERSION=x.x.x
 
-FROM golang:${GOLANG_VERSION} AS build
+FROM nvidia/cuda:${CUDA_VERSION}-base-${BASE_DIST}
```
cdesiniotis, line 18:
```suggestion
FROM nvcr.io/nvidia/cuda:${CUDA_VERSION}-base-${BASE_DIST}
```

tariq1890: Thanks. How did it work before?

cdesiniotis: I think the default registry is `docker.io` so it was pulling the base image from there instead of nvcr.io.

deployments/container/Dockerfile.ubi8
```diff
@@ -39,7 +52,6 @@ RUN chmod a+x /usr/bin/kubectl
 
 COPY driver-manager /usr/local/bin
 COPY scripts/vfio-manage /usr/local/bin
-COPY --from=build /work/nvdrain /usr/local/bin
```
cdesiniotis, line 42: Why was this removed? We still need to copy over the nvdrain binary from the build step

tariq1890: I am not using multi-staged builds anymore. I've specified the output dir of the bin in the go build command

deployments/container/Dockerfile.ubi8
tariq1890: ah got it

deployments/container/Dockerfile.ubi8
cdesiniotis: Oh I missed this detail. Why remove the multi-staged build? Including the go installation in our final image will increase the image size.

tariq1890: Ah yes, fair point. I'll revert it back to multi-stage builds

cdesiniotis (approved):

PR #35, opened 2024-07-17
[no-relnote] Only trigger signing jobs on tags

elezar:


shivakunv (approved):
LGTM

.nvidia-ci.yml
```diff
@@ -184,19 +184,18 @@ release:ngc-ubi8:
...
-      DIST: ["CI_COMMIT_TAG", "ubi8"]
+    - DIST: ["ONLY_CI_COMMIT_TAG", "ubi8"]
   rules:
-    - if: '$DIST == "CI_COMMIT_TAG"'
+    - if: $CI_COMMIT_TAG && '$DIST == "ONLY_CI_COMMIT_TAG"'
       variables:
-        IMAGE_TAG: "${SIGN_JOB_NAME}${CI_COMMIT_TAG}"
-    - when: always
+        IMAGE_TAG: "${CI_COMMIT_TAG}"
+    - if: $CI_COMMIT_TAG
```
cdesiniotis, line 199: Note, we have a duplicate `if: $CI_COMMIT_TAG` conditional in the `.sign:ngc` template. We need to make sure this is defined only once.

cdesiniotis (changes requested):

.nvidia-ci.yml
elezar: Could you clarify what you mean by "this" here? It is my understanding that rules in gitlab CI are evaluated top to bottom and triggered on the first match. Rules that follow are not evaluated.

cdesiniotis: Are the rules defined in `sign:ngc` evaluated before the rules defined in `.sign:ngc`? If that is the case then we can resolve this comment.

elezar: If we define `rules` here, these override the `rules` in the base definition. We are NOT extending the rules but replacing them.

PR #74, opened 2025-06-05
Switch to distroless base image

shivakunv:


scripts/vfio-manage
```diff
@@ -95,56 +89,54 @@ unbind_device() {
...
+    gpu=$1
     device_class_file=$(readlink -f "/sys/bus/pci/devices/$gpu/class")
     device_class=$(cat "$device_class_file")
     if [ "$device_class" != "0x030000" ]; then
       echo "NONE"
       return
     fi
 
-    if ls "/sys/bus/pci/devices/$gpu" | grep consumer >& /dev/null; then
-      aux_dev=$(ls "/sys/bus/pci/devices/$gpu" | grep consumer | awk -Fconsumer:pci: '{print $2}')
```
shivakunv, line 129: shellcheck error: 

aux_dev=$(ls "/sys/bus/pci/devices/$gpu" | grep consumer | awk -Fconsumer:pci: '{print $2}')
                ^-- SC2010 (warning): Don't use ls | grep. Use a glob or a for loop with a condition to allow non-alphanumeric filenames.

cdesiniotis: The shellcheck message you shared is a warning, not an error. I would prefer if we reverted this change to reduce the chance of regression.

The shellcheck error that we must address is: `In POSIX sh, >& is undefined`. To address this, let's update `>& /dev/null` to `>/dev/null 2>&1` as I have done in https://github.com/NVIDIA/gpu-operator/pull/1496

scripts/vfio-manage
```diff
@@ -199,7 +191,7 @@ case "${command}" in
     help) options="" ;;
     *) usage ;;
 esac
-if [ $? -ne 0 ]; then
+if ! "$command"; then
```
cdesiniotis, line 194: I am not quite following why this change was made -- the new conditional is checking something different than before. Can you revert this? Note, the shellcheck message about this line is just a stylistic comment.

cdesiniotis (commented):
Left a review of just the `vfio-manage` script since I am making the same changes in https://github.com/NVIDIA/gpu-operator/pull/1496

deployments/container/Dockerfile.ubi9
```diff
@@ -41,19 +41,24 @@ ENV PATH $GOPATH/bin:/usr/local/go/bin:$PATH
...
+    curl -LO https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/${OS_ARCH}/kubectl && \
+    chmod +x ./kubectl
+
+FROM nvcr.io/nvidia/distroless/go:v3.1.8-dev
 
-RUN dnf install -y kmod
+USER 0:0
 
-# Install K8s client to fetch node resources, cluster version and to perform cleanup of pods for driver unload
-RUN OS_ARCH=${TARGETARCH/x86_64/amd64} && OS_ARCH=${OS_ARCH/aarch64/arm64} && curl -o /usr/bin/kubectl -L "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/${OS_ARCH}/kubectl";
```
tariq1890, line 51: This is the recommended download link for the kubectl binary. Let's keep this as-is please

shivakunv: added thanks . 
PTAL

shivakunv:
resolved all comments . 
PTAL @cdesiniotis @tariq1890

driver-manager
```diff
@@ -605,22 +574,23 @@ uninstall_driver() {
                 _exit_failed
             fi
             echo "Attempting node drain"
-            trap '_exit_failed' ERR
-            _drain_k8s_node
-            _cleanup_driver
+            if ! _drain_k8s_node || ! _cleanup_driver; then
+                echo "Failed to uninstall nvidia driver components"
+                _exit_failed
+            fi
```
cdesiniotis, line 580:
```suggestion
            if ! _drain_k8s_node; then
                echo "Failed to drain node"
                _exit_failed
            fi
            if ! _cleanup_driver; then
                echo "Failed to cleanup NVIDIA driver"
                _exit_failed
            fi
```

driver-manager
```diff
@@ -605,22 +574,23 @@ uninstall_driver() {
...
+        if ! _cleanup_driver; then
             if _is_auto_drain_enabled; then
                 echo "Unable to cleanup driver modules, attempting again with node drain..."
-                trap '_exit_failed' ERR
-                _drain_k8s_node
-                _cleanup_driver
+                if ! _drain_k8s_node || ! _cleanup_driver; then
+                    echo "Failed to uninstall nvidia driver components"
+                    _exit_failed
+                fi
```
cdesiniotis, line 593:
```suggestion
            if ! _drain_k8s_node; then
                echo "Failed to drain node"
                _exit_failed
            fi
            if ! _cleanup_driver; then
                echo "Failed to cleanup NVIDIA driver"
                _exit_failed
            fi
```

driver-manager
```diff
@@ -787,8 +753,6 @@ case "${command}" in
     preflight_check) ;;
     *) usage ;;
 esac
-if [ $? -ne 0 ]; then
+if ! "$command"; then
```
cdesiniotis, line 756: This mistakenly changes behavior. Please revert this change.

Prior to this change, the if statement was checking the exit status of the prior case statement. The proposed change does not do that anymore.

driver-manager
```diff
@@ -787,8 +753,6 @@ case "${command}" in
...
     usage
 fi
-
-$command
```
cdesiniotis, line 794: Revert this. See my prior comment.

shivakunv:
> @shivakunv the commit message is still not entirely accurate. Can you change it to `Switch to distroless base image and update bash scripts to POSIX sh`

changed

cdesiniotis (approved):
@shivakunv thanks for the patience on this!

PR #103, opened 2025-10-02
Add shouldSkipUninstall to avoid GPU driver teardown on restart

karthikvetrivel:
This PR is a part of this endeavor: 
>GPU Driver container should avoid re-installing drivers on spurious container restarts 

Here are a list of scenarios and results in `k8s-driver-manager` and `gpu-driver-container `
https://docs.google.com/document/d/1zPZes_BOuFIfi2_FKb8SDv3eH3I6CLeyua-2h2KS2sE/edit?usp=sharing

### DIFF
--- a/cmd/driver-manager/main.go
+++ b/cmd/driver-manager/main.go
@@ -78,6 +78,8 @@ type config struct {
 	gpuDirectRDMAEnabled       bool
 	useHostMofed               bool
 	kubeconfig                 string
+	driverVersion              string
+	forceReinstall             bool
 }
 
 // ComponentState tracks the deployment state of GPU operator components
@@ -209,6 +211,20 @@ func main() {
 			EnvVars:     []string{"KUBECONFIG"},
 			Value:       "",
 		},
+		&cli.StringFlag{
+			Name:        "driver-version",
+			Usage:       "Desired NVIDIA driver version",
+			Destination: &cfg.driverVersion,
+			EnvVars:     []string{"DRIVER_VERSION"},
+			Value:       "",
+		},
+		&cli.BoolFlag{
+			Name:        "force-reinstall",
+			Usage:       "Force driver reinstall regardless of current state",
+			Destination: &cfg.forceReinstall,
+			EnvVars:     []string{"FORCE_REINSTALL"},
+			Value:       false,
+		},
 	}
 
 	app.Commands = []*cli.Command{
@@ -270,6 +286,11 @@ func (dm *DriverManager) uninstallDriver() error {
 		// Wait for pod termination
 		time.Sleep(60 * time.Second)
 		return fmt.Errorf("driver is pre-installed on host")
+	}
+
+	if skip, reason := dm.shouldSkipUninstall(); skip {
+		dm.log.Infof("Skipping driver uninstall: %s", reason)
+		return nil
 	}
 
 	// Fetch current component states
@@ -621,6 +642,68 @@ func (dm *DriverManager) waitForPodsToTerminate() error {
 func (dm *DriverManager) isDriverLoaded() bool {
 	_, err := os.Stat("/sys/module/nvidia/refcnt")
 	return err == nil
+}
+
+func (dm *DriverManager) shouldSkipUninstall() (bool, string) {
+	if dm.config.forceReinstall {
+		dm.log.Info("Force reinstall is enabled, skipping driver uninstall")
+		return false, ""
+	}
+
+	if !dm.isDriverLoaded() {
+		return true, "no NVIDIA driver modules detected"
+	}
+
+	if dm.config.driverVersion == "" {
+		return false, ""
+	}
+
+	version, err := dm.detectCurrentDriverVersion()
+	if err != nil {
+		dm.log.Warnf("Unable to determine installed driver version: %v", err)
+		// If driver is loaded but we can't detect version, skip uninstall to avoid disruption
+		return true, "driver is loaded but version cannot be determined - skipping to avoid disruption"
+	}
+
+	if version == dm.config.driverVersion {
+		return true, "desired version already present"
+	}
+
+	dm.log.Infof("Installed driver version %s does not match desired %s", version, dm.config.driverVersion)
+	return false, ""
+}
+
+func (dm *DriverManager) detectCurrentDriverVersion() (string, error) {
+	baseCtx := dm.ctx
+	if baseCtx == nil {
+		baseCtx = context.Background()
+	}
+
+	ctx, cancel := context.WithTimeout(baseCtx, 10*time.Second)
+	defer cancel()
+
+	// Try chroot to /run/nvidia/driver for containerized driver
+	cmd := exec.CommandContext(ctx, "chroot", "/run/nvidia/driver", "modinfo", "-F", "version", "nvidia")
+	cmd.Env = append(os.Environ(), "LC_ALL=C")
+	cmdOutput, chrootErr := cmd.Output()
+	if chrootErr == nil {
+		version := strings.TrimSpace(string(cmdOutput))
+		if version != "" {
+			dm.log.Infof("Driver version detected via chroot: %s", version)
+			return version, nil
+		}
+	}
+
+	// Second try to read from /sys/module/nvidia/version if available
+	if versionData, err := os.ReadFile("/sys/module/nvidia/version"); err == nil {
+		version := strings.TrimSpace(string(versionData))
+		if version != "" {
+			dm.log.Infof("Driver version detected from /sys/module/nvidia/version: %s", version)
+			return version, nil
+		}
+	}
+
+	return "", fmt.Errorf("all version detection methods failed: chroot: %v", chrootErr)
 }
 
 func (dm *DriverManager) isNouveauLoaded() bool {


cmd/driver-manager/main.go
```diff
@@ -272,6 +288,11 @@ func (dm *DriverManager) uninstallDriver() error {
 		return fmt.Errorf("driver is pre-installed on host")
 	}
 
+	if skip, reason := dm.shouldSkipUninstall(); skip {
+		dm.log.Infof("Skipping driver uninstall: %s", reason)
+		return nil
```
cdesiniotis, line 293: Returning early here can be problematic. Even if no `nvidia` modules are loaded, there are other operations that get executed below, like handling the vfio-pci driver unloading and waiting for MOFED, that are still relevant. We may want to refactor the control flow here.

cmd/driver-manager/main.go
```diff
@@ -623,6 +644,68 @@ func (dm *DriverManager) isDriverLoaded() bool {
 	return err == nil
 }
 
+func (dm *DriverManager) shouldSkipUninstall() (bool, string) {
+	if dm.config.forceReinstall {
+		dm.log.Info("Force reinstall is enabled, skipping driver uninstall")
```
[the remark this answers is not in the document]
karthikvetrivel: You're right, good catch.

cmd/driver-manager/main.go
karthikvetrivel: I don't exactly remember but the DS couldn't achieve a running state. I think this makes sense though, forcing those “stateless” steps (e.g. vfio-pci) even during the skip path left the GPUs unbound from the driver, so the driver pod never recovered and reverting restored normal behavior.

cdesiniotis: Ah yes -- performing the `unbind` is problematic as the driver container (as you have implemented it currently) will not re-load the module when it starts again, doh!

cmd/driver-manager/main.go
```diff
@@ -40,14 +41,25 @@ import (
...
-	driverRoot          = "/run/nvidia/driver"
-	driverPIDFile       = "/run/nvidia/nvidia-driver.pid"
-	operatorNamespace   = "gpu-operator"
-	pausedStr           = "paused-for-driver-upgrade"
-	defaultDrainTimeout = time.Second * 0
-	defaultGracePeriod  = 5 * time.Minute
+	driverRoot            = "/run/nvidia/driver"
+	driverPIDFile         = "/run/nvidia/nvidia-driver.pid"
+	driverConfigStateFile = "/run/nvidia/driver-config.state"
+	operatorNamespace     = "gpu-operator"
```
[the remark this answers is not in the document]
karthikvetrivel: This definition was in the original version, not sure why it is in the diff. If `OPERATOR_NAMESPACE` is set, that value is used. Otherwise, it uses the `Value` field as a default, which is what we use this variable for.

cmd/driver-manager/main.go
```diff
@@ -629,6 +665,115 @@ func (dm *DriverManager) isDriverLoaded() bool {
...
+	return true, "driver configuration changed"
+}
+
+func (dm *DriverManager) shouldSkipUninstall() (bool, string) {
+	if dm.config.forceReinstall {
+		dm.log.Info("Force reinstall is enabled, proceeding with driver uninstall")
+		return false, ""
+	}
+
+	if !dm.isDriverLoaded() {
```
shivamerla: If the driver is not loaded, then we can skip uninstall (i.e return true here)

cmd/driver-manager/main.go
```diff
@@ -40,15 +41,23 @@ import (
...
+	pausedStr             = "paused-for-driver-upgrade"
+	defaultDrainTimeout   = time.Second * 0
+	defaultGracePeriod    = 5 * time.Minute
 
 	nvidiaDomainPrefix = "nvidia.com"
 
+	nvidiaModuleConfigFile        = "/drivers/nvidia.conf"
+	nvidiaUVMModuleConfigFile     = "/drivers/nvidia-uvm.conf"
+	nvidiaModsetModuleConfigFile  = "/drivers/nvidia-modeset.conf"
+	nvidiaPeermemModuleConfigFile = "/drivers/nvidia-peermem.conf"
```
cdesiniotis: Question -- I don't see an extra volume mount introduced in https://github.com/NVIDIA/gpu-operator/pull/1746 for making sure the `k8s-driver-manager` init container has access to these files. Am I missing something? Since we already have access to `/run/nvidia/driver` does it make sense to get these files from there?

karthikvetrivel: Great catch, Chris! I naively assumed that the init container had this volume mount. 

Updated the paths to read from `/run/nvidia/driver/drivers/*.conf` instead. One thing to note is that these files are only accessible there after the driver has run at least once (as opposed to if we mount the ConfigMap as a volume in the init container). However, that should be fine because `readModuleParams()` handles missing files and `driverModuleBuildNeeded` should return true because no previous driver config should exist.

cmd/driver-manager/main.go
```diff
@@ -629,6 +674,89 @@ func (dm *DriverManager) isDriverLoaded() bool {
...
+}
+
+func (dm *DriverManager) shouldSkipUninstall() bool {
+	if dm.config.forceReinstall {
+		dm.log.Info("Force reinstall is enabled, proceeding with driver uninstall")
+		return false
+	}
+
+	// Only skip uninstall if driver IS loaded AND config matches (fast path optimization)
+	if dm.isDriverLoaded() && !dm.driverModuleBuildNeeded() {
```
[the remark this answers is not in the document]
karthikvetrivel: Updated.

cmd/driver-manager/main.go
```diff
@@ -629,6 +674,92 @@ func (dm *DriverManager) isDriverLoaded() bool {
...
+	}
+
+	// Read module parameters from conf files
+	nvidiaParams := readModuleParams(nvidiaModuleConfigFile)
+	nvidiaUVMParams := readModuleParams(nvidiaUVMModuleConfigFile)
+	nvidiaModeset := readModuleParams(nvidiaModesetModuleConfigFile)
+	nvidiaPeermem := readModuleParams(nvidiaPeermemModuleConfigFile)
+
+	var config strings.Builder
+	config.WriteString(fmt.Sprintf("DRIVER_VERSION=%s\n", dm.config.driverVersion))
```
tariq1890: Instead of relying on multiple `fmt.Sprintf` statements each with a `\n`, you could use a multiline string. One way is  using go template. You could render this template through a struct which represents the driver config state. This would make it more "programmatic" IMO as we establish a scheme for our persisted driver config state

tariq1890: Actually, the templated string and struct could be done in a follow-up.

I see that `fmt.Sprintf` supports interpolation of multi-line backticked strings. https://go.dev/play/p/eoQ3DWKXTtU

Let's use multi-line back-tick strings instead

karthikvetrivel: Updated.

cmd/driver-manager/main.go
```diff
@@ -629,6 +674,102 @@ func (dm *DriverManager) isDriverLoaded() bool {
...
+	data, err := os.ReadFile(filepath)
+	if err != nil {
+		return ""
+	}
+	// Convert newlines to spaces to match bash implementation
+	return strings.ReplaceAll(strings.TrimSpace(string(data)), "\n", " ")
+}
+
+// isDesiredDriverLoaded checks if the driver is loaded and the configuration matches
+func (dm *DriverManager) isDesiredDriverLoaded() bool {
```
tariq1890: Can we rename the method to `shouldUpdateDriverConfig`? If this method returns true, we go don't skip the `Uninstall`. If it's false, we trigger the fast path

cmd/driver-manager/main.go
```diff
@@ -629,6 +674,102 @@ func (dm *DriverManager) isDriverLoaded() bool {
...
+DRIVER_TYPE=%s
+KERNEL_VERSION=%s
+GPU_DIRECT_RDMA_ENABLED=%v
+USE_HOST_MOFED=%v
+KERNEL_MODULE_TYPE=%s
+NVIDIA_MODULE_PARAMS=%s
+NVIDIA_UVM_MODULE_PARAMS=%s
+NVIDIA_MODESET_MODULE_PARAMS=%s
+NVIDIA_PEERMEM_MODULE_PARAMS=%s
+`
```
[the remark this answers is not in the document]
karthikvetrivel: There might be a few difficulties in this approach. 

1) The operator cannot know KERNEL_VERSION at render time, right? What if nodes run heterogeneous kernels that update independently.

2) The container must check `/sys/module/nvidia/refcnt` to verify actual kernel state, which I think that operator cannot do since it lacks access to each node's `/sys` filesystem (fact-check me on this). Even with operator digests, containers still need config-building logic to write state files post-install, so we wouldn't see too much code-savings. 

I think the duplication is worth having the logic live in the containers. Each container can independently validates its own node's actual state (kernel version via uname -r, module state via /sys/module/nvidia/refcnt) without depending on operator coordination. 

Let me know if I misunderstood.

cmd/driver-manager/main.go
```diff
@@ -40,15 +41,23 @@ import (
...
 const (
-	driverRoot          = "/run/nvidia/driver"
-	driverPIDFile       = "/run/nvidia/nvidia-driver.pid"
-	operatorNamespace   = "gpu-operator"
-	pausedStr           = "paused-for-driver-upgrade"
-	defaultDrainTimeout = time.Second * 0
-	defaultGracePeriod  = 5 * time.Minute
+	driverRoot            = "/run/nvidia/driver"
+	driverPIDFile         = "/run/nvidia/nvidia-driver.pid"
+	driverConfigStateFile = "/run/nvidia/driver-config.state"
```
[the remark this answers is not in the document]
karthikvetrivel: Updated.

cmd/driver-manager/main.go
```diff
@@ -629,6 +674,102 @@ func (dm *DriverManager) isDriverLoaded() bool {
...
+}
+
+// readModuleParams reads a module parameter config file and returns its contents as a single-line space-separated string
+func readModuleParams(filepath string) string {
+	data, err := os.ReadFile(filepath)
+	if err != nil {
+		return ""
+	}
+	// Convert newlines to spaces to match bash implementation
+	return strings.ReplaceAll(strings.TrimSpace(string(data)), "\n", " ")
```
tariq1890: In this PR you seem to have two ways to sanitise string output

1. `strings.ReplaceAll(strings.TrimSpace(string(data)), "\n", " ")`
2. `nullIdx := bytes.IndexByte(release, 0)
	return string(release[:nullIdx])`

I suggest using a single approach instead. I believe if you replace the above with the following, it should get you the output you want

```
strings.Trim(string(data), " \r\n\t\u0000\uffff")
```

cmd/driver-manager/main.go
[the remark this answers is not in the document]
tariq1890: I like the idea of decoupling the driver state management from the driver manager, but I am not sure if we should do away with persisting the `KERNEL_VERSION` in the state. How would we account for scenarios like Live kernel patching or similar mechanisms where the kernel maybe updated without node reboot?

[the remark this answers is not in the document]
cdesiniotis: I am not very familiar with live kernel patching, but from scanning https://docs.kernel.org/livepatch/livepatch.html I would not expect the actual kernel version string (as reported by `uname`) to change -- live patching allows you to patch certain functions in kernel code without regenerating the entire kernel image. SUSE has some [docs](https://documentation.suse.com/sled/15-SP6/html/SLED-all/cha-klp.html#sec-klp-overview) that highlight this as well:
> To determine the kernel patching status, use the klp -v patches command. The uname command's output does not change for patched kernels.

With that said, even if we determine we need to persist `KERNEL_VERSION` in the state file, I think my proposal could still work. The state file could be two lines:
```
<digest of current config>
<kernel version>
```
The k8s-driver-manager would build this file and compare it with the existing file at `/run/nvidia/nvidia-driver.state`.

cmd/driver-manager/main.go
karthikvetrivel: You're correct - the init container reads module params from `/run/nvidia/driver/drivers/*.conf`, which doesn't exist yet when the init container runs, so adding a ConfigMap would result in both stored and current configs reading as empty strings. This will trigger fast-track unfortunately. 

Two possible fixes: (1) use the operator-computed digest approach that includes ConfigMap resourceVersion in the hash, or (2) mount the kernel module ConfigMap into the init container as well (at /drivers) so it can read the actual parameter files before the main container starts.

cmd/driver-manager/main.go
```diff
@@ -629,6 +673,100 @@ func (dm *DriverManager) isDriverLoaded() bool {
...
 }
 
+// getKernelVersion returns the current kernel version
+func getKernelVersion() string {
+	var utsname unix.Utsname
+	if err := unix.Uname(&utsname); err != nil {
+		return ""
+	}
+
+	return strings.Trim(string(utsname.Release[:]), " \r\n\t\u0000\uffff")
```
tariq1890: Let's move this to a private method and call that method from `getKernelVersion` and `readModuleParams` method.

Let's name this method `SanitizeString` perhaps

PR #122, opened 2025-11-04
Build multiarch images on native GitHub runners

rajathagasthya:
This commit makes the following changes:

1. Builds multiarch images for non-release commits/PRs, in addition to released versions.
2. Runs the docker build for an arch on the respective GitHub runner (e.g. a linux/amd64 docker image will be built on a linux/amd64 runner). This native build reduces build times due to not requiring emulation.
3. Removes explicit use of buildx for docker build. Buildx is now the default builder in Docker. Also removes the related QEMU setup.


ArangoGutierrez (commented):
FIrst pass, just a NIT comment

cdesiniotis (approved):

Issue #118, opened 2025-10-21
Add retry logic when updating node labels


cdesiniotis:
After uninstalling a previously installed GPU driver, the k8s-driver-manager attempts to reschedule all of the GPU Operator operands by updating node labels. If this operation fails, we do not treat this as a fatal error and instead just emit a warning message: https://github.com/NVIDIA/k8s-driver-manager/blob/a837602b5d1a5c3a62ea2166a19d71fea2cf0984/cmd/driver-manager/main.go#L369. Because of this, it is possible for the GPU Operator state machine to enter a deadlock where the new driver daemonset has come up but all the operands never get rescheduled (if a transient error occurs when attempting to update node labels).

At minimum, we should treat this as a fatal error and have the k8s-driver-manager exit with a non-zero exit code. This way, the k8s-driver-manager will restart and attempt to update the node labels again. We could also consider adding restart logic to the code itself.

myeolenv:
Closing this issue as it is addressed with this PR https://github.com/NVIDIA/k8s-driver-manager/pull/139

PR #141, opened 2025-12-09
Add Error check for Node Label Failure

JunAr7112:
This MR is related to this issue https://github.com/NVIDIA/k8s-driver-manager/issues/118. The goal is to ensure that when the k8s-driver-manager attempts to reschedule the GPU Operator operands, if it fails, we treat this as a fatal error. That way we wouldn't be in a state where the driver manager is ready but the other operands have not come up.

In a previous [MR](https://github.com/NVIDIA/k8s-driver-manager/pull/139) where this was discussed, we added logic to retry the updateNodeLabels() operations in case it ran into errors. That way the k8s-driver-manager will attempt to reschedule the operands to run if the initial operation fails. This MR ensures that if it continues to error after several retries, we treat this as fatal as discussed in the issue rather than just have a warning.

### DIFF
--- a/cmd/driver-manager/main.go
+++ b/cmd/driver-manager/main.go
@@ -395,7 +395,7 @@ func (dm *DriverManager) uninstallDriver() error {
 	}
 
 	if err := dm.rescheduleGPUOperatorComponents(); err != nil {
-		dm.log.Warnf("Failed to reschedule GPU operator components: %v", err)
+		return fmt.Errorf("failed to reschedule GPU operator components: %w", err)
 	}
 
 	// Handle nouveau driver


cmd/driver-manager/main.go
```diff
@@ -395,7 +395,7 @@ func (dm *DriverManager) uninstallDriver() error {
 	}
 
 	if err := dm.rescheduleGPUOperatorComponents(); err != nil {
-		dm.log.Warnf("Failed to reschedule GPU operator components: %v", err)
+		return fmt.Errorf("failed to reschedule GPU operator components: %w", err)
```
rajathagasthya, line 398: This changes behavior of the uninstall operation from best effort rescheduling of GPU Operator components. I'm not sure this change is warranted, but I'd like @cdesiniotis and @tariq1890 to weigh in.

JunAr7112: I think in most cases, the retry logic we merged earlier should resolve this issue, this logic is essentially a double check on the edge case that we won't end up in a state where k8s-driver-manager is up without the other other operands.

cdesiniotis: I think this change is warranted given what I described in https://github.com/NVIDIA/k8s-driver-manager/issues/118. If we fail to reschedule the operands then the gpu-operator state machine will not recover on its own.

rajathagasthya (approved):

cdesiniotis:
@JunAr7112 can you please sign your commit?

JunAr7112:
> @JunAr7112 can you please sign your commit?

Alright I signed it

PR #146, opened 2026-01-07
Skip vfio-pci unbind when GPUs already bound in VFIO mode

karthikvetrivel:
Relevant PR: https://github.com/NVIDIA/gpu-operator/pull/2079

## Description

Prevents unnecessary GPU unbind/rebind operations during rolling updates of the vfio-manager DaemonSet. Currently, k8s-driver-manager unconditionally unbinds all GPUs from vfio-pci on startup, even when the desired state is already vfio-pci. This disrupts active VM workloads using GPU passthrough (KubeVirt, Kata Containers).

### Solution

Check for `GPU_WORKLOAD_CONFIG` environment variable to determine if the pod is running in VFIO workload mode. When set to `vm-passthrough`, skip the vfio-pci unbind operation since GPUs should remain bound to vfio-pci.

### Behavior by DaemonSet Context

| DaemonSet | `GPU_WORKLOAD_CONFIG` | Unbind vfio-pci? | Reason |
|-----------|----------------------|------------------|--------|
| nvidia-driver | _(not set)_ | Yes | Prepares GPUs for NVIDIA kernel driver |
| vfio-manager | `vm-passthrough` |  No | GPUs should remain bound to vfio-pci |
| vgpu-manager | _(not set)_ | Yes | Prepares GPUs for vGPU manager |

### DIFF
--- a/cmd/driver-manager/main.go
+++ b/cmd/driver-manager/main.go
@@ -796,9 +796,13 @@ func (dm *DriverManager) unmountRootfs() error {
 }
 
 // If vfio-pci driver is in use, ensure we unbind it from all devices.
-// If vfio-pci driver is not in use, and we have reached this point, all devices will not be bound to any driver,
-// so the below unbind operation will be a no-op.
+// In VFIO workload mode (vm-passthrough), skip unneccsary unbind.
 func (dm *DriverManager) unbindVfioPCI() error {
+	if dm.isVFIOWorkloadConfig() {
+		dm.log.Info("VFIO workload mode, skipping vfio-pci unbind")
+		return nil
+	}
+
 	dm.log.Info("Unbinding vfio-pci driver from all devices")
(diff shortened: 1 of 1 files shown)


cmd/driver-manager/main.go
```diff
@@ -874,3 +888,43 @@ func (dm *DriverManager) cleanupOnFailure() {
...
 }
+
+func (dm *DriverManager) isVFIOWorkloadConfig() bool {
+	workloadConfig, err := dm.kubeClient.GetNodeLabelValue(dm.config.nodeName, gpuWorkloadConfigLabelKey)
+	if err != nil {
+		dm.log.Warnf("Failed to get workload config label: %v", err)
+		return false
+	}
+
+	return strings.HasPrefix(workloadConfig, "vm-")
```
cdesiniotis: We only want to return true when the workload config is `vm-passthrough` (not `vm-vgpu`)

PR #177, opened 2026-04-27
add support for resetting third party gpu client pods

tariq1890:
### Problem

This commit introduces support for bouncing gpu client pods that are not directly managed by the gpu-operator. Currently, we have users who deploy gpu client applications like the DRA driver, NVSentinel etc along with the gpu-operator. Since these applications are not a part of the static list of operands that the k8s-driver-manager resets during driver upgrades, users have to run into issues with driver upgrades as unmanaged gpu client applications like the DRA driver and NVSentinel aren't bounced at the same time as the gpu-operator operands. 

### Solution

We introduce a new annotation (`nvidia.com/gpu.client: "true"`)to designate a pod that may need to be reset by the k8s-driver-manager during driver upgrades. This way, cluster admins can apply the annotation on any of the pods that act as gpu clients and need to be reset when a driver upgrade needs to happen.

1. The driver-manager looks for pods on the node it's running on with the `nvidia.com/gpu.client: "true"`
2. If pods are found, the driver-manager applies a `nvidia.com/gpu:NoSchedule` taint.
3. Since the taint has been applied, the deleted pods won't be rescheduled on the node
4. The pods from step 1 are then deleted. The driver-manager waits till these pods are terminated
5. The driver-manager performs the unloading of kernel modules and the unbinding of the `/run/nvidia/driver` dir (existing behaviour)
6. The driver-manager removes the taint from step 2

### DIFF
--- a/cmd/driver-manager/main.go
+++ b/cmd/driver-manager/main.go
@@ -33,6 +33,7 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli/v2"
 	"golang.org/x/sys/unix"
+	corev1 "k8s.io/api/core/v1"
 
 	"github.com/NVIDIA/k8s-driver-manager/internal/info"
 	kube "github.com/NVIDIA/k8s-driver-manager/internal/kubernetes"
@@ -62,6 +63,17 @@ const (
 	nvidiaSandboxValidatorDeployLabel    = nvidiaDomainPrefix + "/" + "gpu.deploy.sandbox-validator"
 	nvidiaSandboxDevicePluginDeployLabel = nvidiaDomainPrefix + "/" + "gpu.deploy.sandbox-device-plugin"
 	nvidiaVGPUDeviceManagerDeployLabel   = nvidiaDomainPrefix + "/" + "gpu.deploy.vgpu-device-manager"
+
+	nvidiaGPUClientAnnotation = nvidiaDomainPrefix + "/" + "gpu.client"
+	nvidiaTaintKey            = nvidiaDomainPrefix + "/" + "gpu-driver-update"
+)
+
+var (
+	driverManagerTaint = corev1.Taint{
+		Key:    nvidiaTaintKey,
+		Value:  "k8s-driver-manager",
+		Effect: corev1.TaintEffectNoSchedule,
+	}
 )
 
 // Configuration holds all the configuration from environment variables
@@ -289,6 +301,15 @@ func (dm *DriverManager) uninstallDriver() error {
 	if err := dm.fetchAutoUpgradeAnnotation(); err != nil {
 		return fmt.Errorf("failed to fetch auto upgrade annotation: %w", err)
 	}
+
+	if err := dm.kubeClient.TaintNode(driverManagerTaint, dm.config.nodeName); err != nil {
+		return fmt.Errorf("failed to taint Node %q: %w", dm.config.nodeName, err)
+	}
+	defer func() {
+		if err := dm.kubeClient.UntaintNode(driverManagerTaint, dm.config.nodeName); err != nil {
+			dm.log.Warnf("Failed to untaint Node %q: %v", dm.config.nodeName, err)
+		}
+	}()
 
 	// Always evict all GPU operator components across a driver restart
 	if err := dm.evictAllGPUOperatorComponents(); err != nil {
@@ -650,6 +671,27 @@ func (dm *DriverManager) waitForPodsToTerminate() error {
 		}
 	}
 
+	dm.log.Info("Checking if there are additional gpu clients that need to be restarted...")
+
+	gpuClientPods, err := dm.kubeClient.ListPodsOnNodeWithAnnotation(nvidiaGPUClientAnnotation, nodeName)
+	if err != nil {
+		dm.log.Errorf("Failed to list the gpu client pods: %v", err)
+		return err
+	}
+
+	if len(gpuClientPods) > 0 {
+		dm.log.Info("Found extra gpu-client pods that need to be reset")
+
+		for _, pod := range gpuClientPods {
+			dm.log.Infof("Waiting for pod %s/%s to shutdown", pod.Namespace, pod.Name)
+
+			if err = dm.kubeClient.EvictPodWithWait(pod.Name, pod.Namespace, defaultGracePeriod); err != nil {
+				dm.log.Errorf("Failed while evicting pod %s/%s and waiting for shutdown: %v", pod.Namespace, pod.Name, err)
+				return err
+			}
+		}
+	}
+
 	return nil
 }
 
@@ -958,5 +1000,10 @@ func (dm *DriverManager) cleanupOnFailure() {
 
 	if err := dm.rescheduleGPUOperatorComponents(); err != nil {
 		dm.log.Warnf("Failed to reschedule GPU operator components during cleanup: %v", err)
+	}
+
+	// Ensure that the taint previously applied is removed during cleanup
+	if err := dm.kubeClient.UntaintNode(driverManagerTaint, dm.config.nodeName); err != nil {
+		dm.log.Warnf("Failed to untaint Node %q: %v", dm.config.nodeName, err)
 	}
 }
--- a/internal/kubernetes/client.go
+++ b/internal/kubernetes/client.go
@@ -25,7 +25,10 @@ import (
 	"time"
 
 	"github.com/sirupsen/logrus"
+	jsonpatch "gopkg.in/evanphx/json-patch.v4"
 	corev1 "k8s.io/api/core/v1"
+	policyv1 "k8s.io/api/policy/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
@@ -84,7 +87,7 @@ func NewClient(ctx context.Context, kubeconfig string, log *logrus.Logger) (*Cli
 func (c *Client) GetNodeLabelValue(nodeName, label string) (string, error) {
 	node, err := c.clientset.CoreV1().Nodes().Get(c.ctx, nodeName, metav1.GetOptions{})
 	if err != nil {
-		return "", fmt.Errorf("failed to get node %s: %w", nodeName, err)
+		return "", fmt.Errorf("failed to get node %q: %w", nodeName, err)
 	}
 
 	if node.Labels == nil {
@@ -108,22 +111,7 @@ func (c *Client) UpdateNodeLabels(nodeName string, nodeLabels map[string]string)
 		return fmt.Errorf("failed to marshal patch: %w", err)
 	}
 
-	backoff := wait.Backoff{
-		Duration: time.Second,
-		Factor:   2.0,
-		Jitter:   0.2,
-		Steps:    7,
-	}
-
-	return retry.OnError(backoff, func(err error) bool {
-		return true
-	}, func() error {
-		_, err := c.clientset.CoreV1().Nodes().Patch(c.ctx, nodeName, types.StrategicMergePatchType, patchBytes, metav1.PatchOptions{})
-		if err != nil {
-			c.log.Warnf("Failed to update labels on node %s, retrying: %v", nodeName, err)
-		}
-		return err
-	})
+	return c.patchNodeWithRetry(nodeName, patchBytes, types.StrategicMergePatchType)
 }
 
 // GetNodeAnnotationValue returns the annotation value given a node name and annotation key
@@ -144,7 +132,7 @@ func (c *Client) CordonNode(nodeName string) error {
 
 	node, err := c.clientset.CoreV1().Nodes().Get(c.ctx, nodeName, metav1.GetOptions{})
 	if err != nil {
-		return fmt.Errorf("failed to get node %s: %w", nodeName, err)
+		return fmt.Errorf("failed to get node %q: %w", nodeName, err)
 	}
 
 	drainHelper := &drain.Helper{Ctx: c.ctx, Client: c.clientset}
@@ -157,11 +145,57 @@ func (c *Client) UncordonNode(nodeName string) error {
 
 	node, err := c.clientset.CoreV1().Nodes().Get(c.ctx, nodeName, metav1.GetOptions{})
 	if err != nil {
-		return fmt.Errorf("failed to get node %s: %w", nodeName, err)
+		return fmt.Errorf("failed to get node %q: %w", nodeName, err)
 	}
 
 	drainHelper := &drain.Helper{Ctx: c.ctx, Client: c.clientset}
 	return drain.RunCordonOrUncordon(drainHelper, node, false)
+}
+
+// TaintNode applies a taint on a Node marking it as Unschedulable. If the taint already exists, it skips updating
+// the node
+func (c *Client) TaintNode(taint corev1.Taint, nodeName string) error {
+	c.log.Infof("Tainting node %s", nodeName)
+
+	node, err := c.clientset.CoreV1().Nodes().Get(c.ctx, nodeName, metav1.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to get node %q: %w", nodeName, err)
+	}
+
+	original := node.DeepCopy()
+
+	// First, we check if the taint already exists. If yes, we skip the node patch operation
+	for _, nodeTaint := range node.Spec.Taints {
+		if taintsEqual(nodeTaint, taint) {
+			c.log.Infof("Taint has already been applied. Skipping node taint...")
+			return nil
+		}
+	}
+
+	node.Spec.Taints = append(node.Spec.Taints, taint)
+
+	return c.applyNodeDiff(original, node)
+}
+
+// UntaintNode removes the taint on a Node. If the taint doesn't exist, it skips updating the node.
+func (c *Client) UntaintNode(taint corev1.Taint, nodeName string) error {
+	c.log.Infof("Untainting node %s", nodeName)
+
+	node, err := c.clientset.CoreV1().Nodes().Get(c.ctx, nodeName, metav1.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to get node %q: %w", nodeName, err)
+	}
+
+	original := node.DeepCopy()
+
+	for i, nodeTaint := range node.Spec.Taints {
+		if taintsEqual(nodeTaint, taint) {
+			node.Spec.Taints = append(node.Spec.Taints[:i], node.Spec.Taints[i+1:]...)
+			return c.applyNodeDiff(original, node)
+		}
+	}
+	c.log.Infof("Taint not found. Skipping node untaint...")
+	return nil
 }
 
 // WaitForPodTermination will wait for the termination of pods matching labels from the selectorMap on the node with the specified namespace.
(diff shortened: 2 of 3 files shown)


cmd/driver-manager/main.go
```diff
@@ -650,6 +663,27 @@ func (dm *DriverManager) waitForPodsToTerminate() error {
...
+	dm.log.Info("Listing additional gpu-management workloads...")
+
+	gpuManagementPods, err := dm.kubeClient.ListPodsOnNodeWithAnnotation(nvidiaGPUClientAnnotation, nodeName)
+	if err != nil {
+		dm.log.Errorf("Failed to list gpu-management pods: %v", err)
+		return err
+	}
+
+	if len(gpuManagementPods) > 0 {
+		dm.log.Info("Found extra gpu-client pods that need to be reset.")
```
cdesiniotis: nit: here we use `gpu-client pods` while above on L668 we use `gpu-management pods`. Let's decide on one and be consistent.

internal/kubernetes/client.go
```diff
@@ -164,6 +151,55 @@ func (c *Client) UncordonNode(nodeName string) error {
 	return drain.RunCordonOrUncordon(drainHelper, node, false)
 }
 
+// TaintNode applies a taint on a Node given a node name marking it as Unschedulable
+func (c *Client) TaintNode(nodeName string) error {
+	c.log.Infof("Tainting node %s", nodeName)
```
cdesiniotis, line 158: I noticed we are logging this message here and at the call site. Should we only log once?

internal/kubernetes/client.go
```diff
@@ -164,6 +151,55 @@ func (c *Client) UncordonNode(nodeName string) error {
...
+		return fmt.Errorf("failed to get node %s: %w", nodeName, err)
+	}
+
+	original := node.DeepCopy()
+
+	taint := corev1.Taint{
+		Key:    nvidiaTaintKey,
+		Value:  "k8s-driver-manager",
+		Effect: corev1.TaintEffectNoSchedule,
+	}
```
cdesiniotis: nit: thoughts on creating a top-level variable for this, named `driverManagerTaint`, that can be referenced in both the `TaintNode` and `UntaintNode` methods?

internal/kubernetes/client.go
tariq1890: will remove

internal/kubernetes/client.go
tariq1890: Sure, I am not opposed to that

tariq1890:
Review comments addressed. Thanks @cdesiniotis !

cdesiniotis (approved):

cmd/driver-manager/main.go
```diff
@@ -350,6 +362,11 @@ func (dm *DriverManager) uninstallDriver() error {
 
 	// Check if driver is loaded and cleanup if needed
 	if dm.isDriverLoaded() {
+		// Apply Taint on node to prevent pods from being scheduled during a driver unload cycle
+		if err := dm.kubeClient.TaintNode(driverManagerTaint, dm.config.nodeName); err != nil {
```
rajathagasthya: Should we apply the taint earlier before the "gpu client" pods are deleted? Is there a chance their controllers can reschedule pods on to the node between deletion and here?

cmd/driver-manager/main.go
```diff
@@ -368,6 +385,9 @@ func (dm *DriverManager) uninstallDriver() error {
 				return fmt.Errorf("failed to uninstall nvidia driver components: %w", err)
 			}
 		}
+		if err := dm.kubeClient.UntaintNode(driverManagerTaint, dm.config.nodeName); err != nil {
```
rajathagasthya: How about un-tainting in a `defer` block? It's cleaner and safe from panics.

internal/kubernetes/client.go
```diff
@@ -293,3 +354,46 @@ func gpuPodSpecFilter(pod corev1.Pod) bool {
...
+	}, func() error {
+		_, err := c.clientset.CoreV1().Nodes().Patch(c.ctx, nodeName, patchType, patch, metav1.PatchOptions{})
+		if err != nil {
+			c.log.Warnf("Failed to patch node %s, retrying: %v", nodeName, err)
+		}
+		return err
+	})
+}
+
+func taintsEqual(t1 corev1.Taint, t2 corev1.Taint) bool {
```
rajathagasthya, line 409: Can we use `corev1.Taint.MatchTaint` and just inline it?

https://pkg.go.dev/k8s.io/api/core/v1#Taint.MatchTaint

cmd/driver-manager/main.go
```diff
@@ -62,6 +63,17 @@ const (
 	nvidiaSandboxValidatorDeployLabel    = nvidiaDomainPrefix + "/" + "gpu.deploy.sandbox-validator"
 	nvidiaSandboxDevicePluginDeployLabel = nvidiaDomainPrefix + "/" + "gpu.deploy.sandbox-device-plugin"
 	nvidiaVGPUDeviceManagerDeployLabel   = nvidiaDomainPrefix + "/" + "gpu.deploy.vgpu-device-manager"
+
+	nvidiaGPUClientAnnotation = nvidiaDomainPrefix + "/" + "gpu.client"
+	nvidiaTaintKey            = nvidiaDomainPrefix + "/" + "gpu"
```
rajathagasthya: Can we make the taint key a bit more specific like `nvidia.com/gpu-driver-upgrade`? This is generic enough that users could already be using this key is my concern.

internal/kubernetes/client.go
```diff
@@ -164,6 +150,58 @@ func (c *Client) UncordonNode(nodeName string) error {
...
+	c.log.Infof("Tainting node %s", nodeName)
+
+	node, err := c.clientset.CoreV1().Nodes().Get(c.ctx, nodeName, metav1.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to get node %s: %w", nodeName, err)
+	}
+
+	original := node.DeepCopy()
+
+	// First, we check if the taint already exists. If yes, we skip the node patch operation
```
rajathagasthya, line 167: super nit: code below is self explanatory, so this comment is not necessary.

cmd/driver-manager/main.go
tariq1890: I chose `nvidia.com/gpu` as it would not necessitate adding a new toleration to our driver daemonset. If a taint is added and the driver daemonset doesn't tolerate it, the daemonset controller will mark the driver daemonset as mis-scheduled and this leads to the upgrade controller failing the driver upgrade.

cmd/driver-manager/main.go
tariq1890: I considered this initially. But I've chosen to call the `UntaintNode` method from the `cleanupOnFailure` method instead. This method is called in all of the error scenarios of the driver manager and I've updated `UntaintNode` to ensure idempotency

cmd/driver-manager/main.go
tariq1890: Please check now

internal/kubernetes/client.go
tariq1890: Done

internal/kubernetes/client.go
tariq1890: Nice, thanks for suggesting this!

tariq1890: Actually, we may not be able to use this. We also need to check the taint value as we plan on having other entities apply similar taints (like mig-manager) and we need to check the `Taint.Value` to ensure the right taints are being added/removed.

```
// MatchTaint checks if the taint matches taintToMatch. Taints are unique by key:effect,
// if the two taints have same key:effect, regard as they match.
func (t *Taint) MatchTaint(taintToMatch *Taint) bool {
	return t.Key == taintToMatch.Key && t.Effect == taintToMatch.Effect
}
```

rajathagasthya: I see. That makes sense, thanks.

cmd/driver-manager/main.go
```diff
@@ -650,6 +671,27 @@ func (dm *DriverManager) waitForPodsToTerminate() error {
...
+		return err
+	}
+
+	if len(gpuClientPods) > 0 {
+		dm.log.Info("Found extra gpu-client pods that need to be reset")
+
+		for _, pod := range gpuClientPods {
+			dm.log.Infof("Waiting for pod %s/%s to shutdown", pod.Namespace, pod.Name)
+
+			if err = dm.kubeClient.DeletePodWithWait(pod.Name, pod.Namespace, defaultGracePeriod); err != nil {
```
rajathagasthya: I'm wondering if we should use the `Eviction` API here instead of deleting the pods. My reasons are:

1. It respects `PodDisruptionBudgets`. In the future, GPU client pods may have PDBs defined and we should respect it. I presume we will document this annotation as a "contract" to reset any GPU client pods on a driver upgrade cycle. If external users use this annotation, they may have PDBs set for their pods.
2. This removes the need for expanded RBAC in https://github.com/NVIDIA/gpu-operator/pull/2408. The driver-manager's ClusterRole already has `pods/eviction:create` granted. `pods/delete` is quite sweeping for a privileged driver daemonset.
3. When we drain node currently, the drain helper already uses Eviction by default.

Let me know your thoughts.

cmd/driver-manager/main.go
rajathagasthya: Actually, let me revisit this. (Please correct me if my assumptions are wrong)

I noticed the DRA Driver's KubeletPlugin already has a toleration for this taint [here](https://github.com/kubernetes-sigs/dra-driver-nvidia-gpu/blob/14178eb0db1ecb426017d6f0091c3d549e7ef321/deployments/helm/dra-driver-nvidia-gpu/values.yaml#L262-L265). It uses `operator: Exists`, which ignores value during taint matching. So it will be rescheduled on the tainted node after deletion since it tolerates. Are we planning to remove that toleration?

For taints, the uniqueness constraint is `(key, effect)`. So if `nvidia.com/gpu:NoSchedule` taint is already present on the node and we try to add the same taint with `value: k8s-driver-manager` below, the API server rejects the patch as duplicate.

```
$ kubectl taint nodes kind-control-plane nvidia.com/gpu:NoSchedule
node/kind-control-plane tainted
$ kubectl taint nodes kind-control-plane nvidia.com/gpu=k8s-driver-manager:NoSchedule
error: node kind-control-plane already has nvidia.com/gpu taint(s) with same effect(s) and --overwrite is false
```
(We could use `--overwrite`, but we will be possibly overwriting a taint we didn't add)

The value comparison in `taintsEqual` doesn't help avoid this error. Value is differentiated only at the toleration level when `operator: Equal` is used, and we can't be sure clients always use this.

A unique (more specific) taint key resolves these cases cleanly and we don't have to rely on value comparison to differentiate that the driver-manager added this taint. I think there's no downside to this, except adding another toleration in our deployments.

rahulait: Also, there is extendedResourceToleration admission controller which automatically adds taints to nodes if enabled based on the resource request.
https://medium.com/robovision-ai/use-kubernetes-extendedresourcetoleration-to-keep-out-non-gpu-pods-c180f5c5e76d

Since we use "nvidia.com/gpu=X" when requesting GPUs, that controller will automatically add taints to resources requesting gpus and then those pods will tolerate the taint (will still get scheduled on eviction). So we should definitely rename this taint to something else like what Rajath suggested.

tariq1890: Understood, thanks for all the details. I'll update the taint; this means that our driver daemonset will have to be updated to tolerate the new taint here

tariq1890: This is done, please check now

cmd/driver-manager/main.go
tariq1890: Thanks @rajathagasthya, I've updated the code to use the eviction API instead of a direct pod delete.

cmd/driver-manager/main.go
tariq1890: I've added the defer block, please check now

internal/kubernetes/client.go
```diff
@@ -293,3 +376,46 @@ func gpuPodSpecFilter(pod corev1.Pod) bool {
...
+		if err != nil {
+			c.log.Warnf("Failed to patch node %s, retrying: %v", nodeName, err)
+		}
+		return err
+	})
+}
+
+func taintsEqual(t1 corev1.Taint, t2 corev1.Taint) bool {
+	keysEqual := t1.Key == t2.Key
+	valuesEqual := t1.Value == t2.Value
```
rajathagasthya, line 411: Since we're no longer sensitive to value, we can just use `corev1.Taint.MatchTaint`, right?

tariq1890: I believe we still need it. I envision the mig-manager also adding a similar when it performs MIG mode enablement and the toggling of the gpu client

```
        - key: nvidia.com/gpu-driver-update
          value: "k8s-mig-manager"
          effect: NoSchedule
```

rajathagasthya: That sounds okay to me.

rajathagasthya (approved):
LGTM! 🚀

tariq1890:
Requesting review on this [PR](https://github.com/NVIDIA/gpu-operator/pull/2408) as well. I'd like to merge them both at the same time

internal/kubernetes/client.go
```diff
@@ -293,3 +376,46 @@ func gpuPodSpecFilter(pod corev1.Pod) bool {
...
+	backoff := wait.Backoff{
+		Duration: time.Second,
+		Factor:   2.0,
+		Jitter:   0.2,
+		Steps:    7,
+	}
+	return retry.OnError(backoff, func(err error) bool {
+		return true
+	}, func() error {
+		_, err := c.clientset.CoreV1().Nodes().Patch(c.ctx, nodeName, patchType, patch, metav1.PatchOptions{})
```
rahulait, line 401: It looks like [NodeSpec.Taints](https://github.com/NVIDIA/k8s-driver-manager/blob/ae3f46db3d06f6019497b8372d80d990e797ee14/vendor/k8s.io/api/core/v1/types.go#L6174-L6175) is +listType=atomic , so it would still replace the whole spec.taints array. In case it retries for longer duration, the chances of it overwriting a taint added by someone else increases. Should we use real read-modify-write with Nodes().Update() here when tainting/untainting nodes. For rest, we'll use patch(). That way Kubernetes rejects stale resourceVersions, and each retry recomputes from the latest taint list?

tariq1890: I am not too convinced that this really poses a risk. Right now, I lean towards making this a patch operation and ensuring the taint succeeds instead of going with an update that is stricter and will necessitate a retry block. I don't want to over-engineer this too much

internal/kubernetes/client.go
```diff
@@ -164,6 +152,59 @@ func (c *Client) UncordonNode(nodeName string) error {
...
+	}
+
+	original := node.DeepCopy()
+
+	found := false
+	for i, nodeTaint := range node.Spec.Taints {
+		if taintsEqual(nodeTaint, taint) {
+			node.Spec.Taints = append(node.Spec.Taints[:i], node.Spec.Taints[i+1:]...)
+			found = true
+			break
```
rahulait: We can avoid extra var and break here by directly calling `return c.applyNodeDiff(original, node)` here. Outside for loop, we can simply log taint not found and return nil.

internal/kubernetes/client.go
```diff
@@ -164,6 +152,59 @@ func (c *Client) UncordonNode(nodeName string) error {
...
+	return c.applyNodeDiff(original, node)
+}
+
+// UntaintNode removes the taint on a Node. If the taint doesn't exist, it skips updating the node.
+func (c *Client) UntaintNode(taint corev1.Taint, nodeName string) error {
+	c.log.Infof("Untainting node %s", nodeName)
+
+	node, err := c.clientset.CoreV1().Nodes().Get(c.ctx, nodeName, metav1.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to get node %s: %w", nodeName, err)
```
rahulait:
```suggestion
		return fmt.Errorf("failed to get node %q: %w", nodeName, err)
```

internal/kubernetes/client.go
```diff
@@ -164,6 +152,59 @@ func (c *Client) UncordonNode(nodeName string) error {
...
 }
 
+// TaintNode applies a taint on a Node marking it as Unschedulable. If the taint already exists, it skips updating
+// the node
+func (c *Client) TaintNode(taint corev1.Taint, nodeName string) error {
+	c.log.Infof("Tainting node %s", nodeName)
+
+	node, err := c.clientset.CoreV1().Nodes().Get(c.ctx, nodeName, metav1.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to get node %s: %w", nodeName, err)
```
rahulait:
```suggestion
		return fmt.Errorf("failed to get node %q: %w", nodeName, err)
```

cmd/driver-manager/main.go
```diff
@@ -650,6 +671,27 @@ func (dm *DriverManager) waitForPodsToTerminate() error {
...
+	}
+
+	if len(gpuClientPods) > 0 {
+		dm.log.Info("Found extra gpu-client pods that need to be reset")
+
+		for _, pod := range gpuClientPods {
+			dm.log.Infof("Waiting for pod %s/%s to shutdown", pod.Namespace, pod.Name)
+
+			if err = dm.kubeClient.EvictPodWithWait(pod.Name, pod.Namespace, defaultGracePeriod); err != nil {
+				dm.log.Errorf("Failed to wait for pod %s/%s to shutdown: %v", pod.Namespace, pod.Name, err)
```
rahulait:
```suggestion
				dm.log.Errorf("Failed while evicting pod %s/%s and waiting for shutdown: %v", pod.Namespace, pod.Name, err)
```

cmd/driver-manager/main.go
```diff
@@ -290,6 +302,15 @@ func (dm *DriverManager) uninstallDriver() error {
 		return fmt.Errorf("failed to fetch auto upgrade annotation: %w", err)
 	}
 
+	if err := dm.kubeClient.TaintNode(driverManagerTaint, dm.config.nodeName); err != nil {
+		return fmt.Errorf("failed to taint Node %s: %w", dm.config.nodeName, err)
+	}
+	defer func() {
+		if err := dm.kubeClient.UntaintNode(driverManagerTaint, dm.config.nodeName); err != nil {
+			dm.log.Warnf("failed to untaint Node %s: %v", dm.config.nodeName, err)
```
rahulait:
```suggestion
			dm.log.Warnf("Failed to untaint Node %q: %v", dm.config.nodeName, err)
```

cmd/driver-manager/main.go
```diff
@@ -290,6 +302,15 @@ func (dm *DriverManager) uninstallDriver() error {
 		return fmt.Errorf("failed to fetch auto upgrade annotation: %w", err)
 	}
 
+	if err := dm.kubeClient.TaintNode(driverManagerTaint, dm.config.nodeName); err != nil {
+		return fmt.Errorf("failed to taint Node %s: %w", dm.config.nodeName, err)
```
rahulait:
```suggestion
		return fmt.Errorf("failed to taint Node %q: %w", dm.config.nodeName, err)
```

cmd/driver-manager/main.go
```diff
@@ -62,6 +63,17 @@ const (
 	nvidiaSandboxValidatorDeployLabel    = nvidiaDomainPrefix + "/" + "gpu.deploy.sandbox-validator"
 	nvidiaSandboxDevicePluginDeployLabel = nvidiaDomainPrefix + "/" + "gpu.deploy.sandbox-device-plugin"
 	nvidiaVGPUDeviceManagerDeployLabel   = nvidiaDomainPrefix + "/" + "gpu.deploy.vgpu-device-manager"
+
+	nvidiaGPUClientAnnotation = nvidiaDomainPrefix + "/" + "gpu.client"
+	nvidiaTaintKey            = nvidiaDomainPrefix + "/" + "gpu-driver-update"
+)
```
rahulait, line 69: Just a note: I am not sure why in the past we went with "gpu.deploy.*" keys. I don't know if the vision was to keep operations within dot after gpu or it was just random. We could have also used "gpu.update.driver" here. I am fine with either approach, just wanted to call out in case there was some history behind this.

rahulait (approved):
few very minor nit picks, but otherwise LGTM.

cmd/driver-manager/main.go
```diff
@@ -959,4 +1001,9 @@ func (dm *DriverManager) cleanupOnFailure() {
 	if err := dm.rescheduleGPUOperatorComponents(); err != nil {
 		dm.log.Warnf("Failed to reschedule GPU operator components during cleanup: %v", err)
 	}
+
+	// Ensure that the taint previously applied is removed during cleanup
+	if err := dm.kubeClient.UntaintNode(driverManagerTaint, dm.config.nodeName); err != nil {
+		dm.log.Warnf("failed to untaint Node %s: %v", dm.config.nodeName, err)
```
rahulait:
```suggestion
		dm.log.Warnf("Failed to untaint Node %q: %v", dm.config.nodeName, err)
```

internal/kubernetes/client.go
rahulait: Ack.

rahulait:
One thing that is still not clear to me is how the system will perform if eviction fails after 5 mins and keeps failing for a long time (say few days bcoz of a PDB preventing more pods to be evicted or something else). The node will remain tainted for those 5 mins, so no pod (gpu or without gpu) will get scheduled onto that node (unless it has the toleration for specific taint). On cleanup, we'll remove the taint and then retry, so there will be a brief window where other pods can get scheduled onto the node and this will go on until fixed. Maybe thats fine and nobody will have concerns that other workloads are not getting scheduled on those nodes immediately.

cmd/driver-manager/main.go
tariq1890: I am definitely open to better names for the taint here. It was the best I could come up with, so I'd certainly appreciate more suggestions. At the same time, I don't think the taint key naming has to adhere to the label naming conventions

tariq1890:
> One thing that is still not clear to me is how the system will perform if eviction fails after 5 mins and keeps failing for a long time (say few days bcoz of a PDB preventing more pods to be evicted or something else). The node will remain tainted for those 5 mins, so no pod (gpu or without gpu) will get scheduled onto that node (unless it has the toleration for specific taint).

I think it'd be better for the node to be in this tainted state if a gpu client is stuck in Terminating state rather than have the an upgrade to proceed. We want to driver upgrades to block on these gpu clients from here on out after all

> so there will be a brief window where other pods can get scheduled onto the node and this will go on until fixed. 

Yes, but what we would expect to happen here is for those pods to go into CrashLoopBackoff or to block on Toolkit being ready if they have an init-container similar to that of our operands.

internal/kubernetes/client.go
tariq1890: Thanks

tariq1890:
This is now superseded by #205

PR #203, opened 2026-06-24
Drain the DRA kubelet-plugin and validator across driver reloads

karthikvetrivel:
This PR adds the NVIDIA DRA driver `kubelet-plugin` and the DRA validator to the set of GPU clients that the `driver-manager` drains off a node before reloading the driver. When the GPU Operator manages GPUs through Dynamic Resource Allocation (the new DRA CRD path), these pods hold the driver just like the device-plugin, GFD, and DCGM do today, so they must be drained and recreated across a driver reload. Otherwise the `kubelet-plugin` keeps serving CDI specs enumerated against the old driver and GPU claims fail until it is manually restarted.

All new behavior is gated on the `nvidia.com/gpu.deploy.dra-driver` and `nvidia.com/gpu.deploy.dra-validator` node labels, so deployments without the DRA stack are unaffected, mirroring the existing optional components (mig-manager, sandbox-*, vgpu-device-manager). 

Functionally, this PR depends on companion GPU Operator changes that set those labels and add the matching `nodeSelector` to the kubelet-plugin and validator DaemonSets.

## Testing
I tested the following scenarios manually on a 2-node cluster (A100 worker + T4 control-plane) running the GPU Operator DRA stack:

- I deployed the DRA stack and verify the kubelet-plugin and dra-validator are labeled with `nvidia.com/gpu.deploy.dra-driver=true` and `nvidia.com/gpu.deploy.dra-validator=true` and come up on each GPU node.
- Then, I triggered a driver upgrade by bumping the NVIDIADriver version. I verified the driver-manager drains the validator and GFD first, then the kubelet-plugin, reloads the driver, and restores them. Both nodes completed the rolling upgrade with no pods stuck `Terminating`. I verified the kubelet-plugin pods were recreated, re-enumerated against the new driver, and a GPU claim succeeded on each node afterward.
- I verified that drain ordering is required. I found that draining the kubelet-plugin in the same batch as the claim-holders left the validator and GFD stuck `Terminating` (their DRA claims cannot be released once the plugin is gone) and deadlocked the upgrade, so draining the plugin last after them avoids this.

### DIFF
--- a/cmd/driver-manager/main.go
+++ b/cmd/driver-manager/main.go
@@ -63,6 +63,8 @@ const (
 	nvidiaSandboxDevicePluginDeployLabel = nvidiaDomainPrefix + "/" + "gpu.deploy.sandbox-device-plugin"
 	nvidiaVGPUDeviceManagerDeployLabel   = nvidiaDomainPrefix + "/" + "gpu.deploy.vgpu-device-manager"
 	nvidiaGPUClientDeployLabel           = nvidiaDomainPrefix + "/" + "gpu.deploy.client"
+	nvidiaDRADriverDeployLabel           = nvidiaDomainPrefix + "/" + "gpu.deploy.dra-driver"
+	nvidiaDRAValidatorDeployLabel        = nvidiaDomainPrefix + "/" + "gpu.deploy.dra-validator"
 )
 
 // Configuration holds all the configuration from environment variables
@@ -95,6 +97,8 @@ type componentState struct {
 	sandboxValidatorDeployed    string
 	sandboxPluginDeployed       string
 	vgpuDeviceManagerDeployed   string
+	draDriverDeployed           string
+	draValidatorDeployed        string
 	customOperandNodeLabelValue string
 	gpuClientsDeployed          string
 	autoUpgradePolicyEnabled    string
@@ -292,7 +296,10 @@ func (dm *DriverManager) uninstallDriver() error {
 		return fmt.Errorf("failed to fetch auto upgrade annotation: %w", err)
 	}
 
-	// Always evict all GPU operator components across a driver restart
+	// Always evict all GPU operator components across a driver restart. The DRA
+	// kubelet-plugin is the exception: it services NodeUnprepareResources for the
+	// claim-holders evicted here (e.g. dra-validator), so it must outlive them and
+	// is drained separately afterwards (see evictKubeletPlugin).
 	if err := dm.evictAllGPUOperatorComponents(); err != nil {
 		dm.log.Error("Failed to evict GPU operator components, attempting cleanup")
 		dm.cleanupOnFailure()
@@ -301,6 +308,17 @@ func (dm *DriverManager) uninstallDriver() error {
 
 	if dm.shouldSkipUninstall() {
 		dm.log.Info("The NVIDIA driver is already loaded with the desired version and configuration, skipping the uninstallation of the driver in an attempt to not disrupt running workloads")
+
+		// The DRA kubelet-plugin bind-mounts the previous driver container's rootfs.
+		// That bind does not track the host mount point, so the replacement driver
+		// container's rootfs never appears inside the running plugin and, once the
+		// stale rootfs is unmounted below, its submounts (e.g. dev/) vanish from the
+		// plugin's view, leaving NodePrepareResources unable to build CDI specs.
+		// Restart the plugin across the rotation so it re-binds the new rootfs.
+		if err := dm.evictKubeletPlugin(); err != nil {
+			dm.cleanupOnFailure()
+			return fmt.Errorf("failed to evict DRA kubelet-plugin: %w", err)
+		}
 
 		// Clean up stale artifacts from previous container before rescheduling operands
 		dm.log.Info("Cleaning up stale mounts and state files...")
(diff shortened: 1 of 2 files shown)


tariq1890 (approved):

cmd/driver-manager/main.go
```diff
@@ -343,13 +347,17 @@ func (dm *DriverManager) uninstallDriver() error {
...
-				dm.cleanupOnFailure()
-				return fmt.Errorf("failed to cleanup NVIDIA driver: %w", err)
-			}
 		}
 	}
 
+	// Drain the kubelet-plugin after the claim-holders and GPU workloads are gone but
+	// before the driver is unloaded, so the kubelet can still reach it to release their
+	// DRA claims.
+	if err := dm.evictKubeletPlugin(); err != nil {
```
rahulait, line 387: Based on codex:

> evictKubeletPlugin() runs unconditionally after the optional GPU-pod eviction block:
> GPU workload eviction occurs only when ENABLE_GPU_POD_EVICTION=true.
> But the DRA kubelet-plugin is always paused immediately afterward.
> If GPU-pod eviction is disabled and automatic node drain is enabled, DRA claim-holder workloads are still running when the plugin is removed.
> Driver unload then fails because those workloads still use the driver. The fallback node drain deletes them only after the kubelet-plugin is gone, so their DRA claims cannot be unprepared. This is the deadlock the PR description correctly says it must avoid.
> The claimed ordering is therefore only valid in the default “GPU-pod eviction enabled” path, not for all supported drain configurations.
> 
> The flow must guarantee:
> DRA claim-holders drained
> → kubelet-plugin drained
> → driver unloaded
> 
> For the GPU-pod eviction disabled + auto-drain enabled path, run the node drain before evictKubeletPlugin, or defer plugin eviction until a claim-holder drain has completed successfully. Do not rely on the post-unload-failure fallback.
> 

Can you review this?

karthikvetrivel: I believe this is valid, thanks for flagging. Should be fixed, just needed to change the condition.

cmd/driver-manager/main.go
rajathagasthya, line 693: This needs `nvidia-dcgm-dra` and `nvidia-dcgm-exporter-dra` also, right?

cdesiniotis: nit: can we mention that the DRA Driver kubelet plugin is the exception here (and why)

cdesiniotis, line 309: Question -- should we be restarting the DRA driver kubelet plugin / validator in the fast path case? I am assuming the kubelet plugin needs the new driver container's root fs mounted.

karthikvetrivel: Right, added. Note the -dra DaemonSets share the same gpu.deploy.dcgm* node-selector labels, so the eviction flip already covered them. 

We only need the app labels for the termination wait.

karthikvetrivel: The dra-validator already restarts in the fast path (it's evicted in the common batch before the skip-uninstall check and rescheduled after). 

I don't think we should restart the kubelet plugin since it mounts the driver install dir with `mountPropagation: HostToContainer`, so the new driver container's rootfs mount propagates into the running plugin. Does that sound right?

cmd/driver-manager/main.go
karthikvetrivel: Addressed.

cmd/driver-manager/main.go
cdesiniotis: I am not entirely certain what the behavior is, so best to try this out.

karthikvetrivel: Updated.

rahulait (approved):

rahulait:
I saw this additional flag raised:

> The remaining safety gap is that the kubelet plugin is always evicted once DRA is enabled, even when the GPU Operator auto-upgrade policy has disabled driver-manager’s own workload eviction. The code assumes the upgrade controller has already drained every DRA workload. Add a pre-plugin check that no Pod on the node still holds an allocation from gpu.nvidia.com; otherwise leave the plugin running and retry/fail clearly.
That guard is needed to prevent a remaining DRA workload from becoming stuck in termination when it later needs NodeUnprepareResources.

cmd/driver-manager/main.go
karthikvetrivel: Updated to restart DRA driver kubelet plugin.

karthikvetrivel:
@rahulait Let me try to reproduce the issue you flagged.

karthikvetrivel:
@rahulait Confirmed that the bug is real and fixed it.  Before draining the kubelet-plugin, driver-manager now lists the node's pods and aborts if any *non-terminal* pod still holds a gpu.nvidia.com claim (w/ retry).

rahulait:
Thanks @karthikvetrivel , LGTM

rahulait (approved):

cdesiniotis (approved):

internal/kubernetes/client.go
```diff
@@ -218,6 +218,38 @@ func (c *Client) WaitForPodsWithNodeSelector(nodeName, nodeSelectorKey string, t
 	})
 }
 
+// CheckNoGPUResourceClaimHolders returns an error naming the pods on the node which still
+// hold a ResourceClaim allocated by the NVIDIA GPU DRA driver, i.e. the pods the kubelet
+// still needs the DRA kubelet-plugin to unprepare. Pods in a terminal phase are excluded,
+// since the kubelet has unprepared their claims by then.
+func (c *Client) CheckNoGPUResourceClaimHolders(nodeName string) error {
```
tariq1890: It's better to have this method return the list of pods holding GPUResourceClaims. We can check if this list is empty or not, and throw an error at the call-site. 

The decision to throw an error based on a non-empty list of gpu resource claim holders is out of scope for this package (`internal/kubernetes/client`)

karthikvetrivel: Amended and pushed.

tariq1890 (approved):

PR #204, opened 2026-06-30
Add DRA support for GPU pod eviction during driver upgrades

karthikvetrivel:
The driver-manager init container evicts GPU pods from a node before reloading the NVIDIA GPU driver, but today it only detects pods that request GPUs via device-plugin resources (nvidia.com/gpu, nvidia.com/mig-*). This PR extends the pod filter to also detect pods using GPUs through Dynamic Resource Allocation (DRA): for each pod on the node, it resolves the referenced ResourceClaim(s) and checks whether any is allocated by the NVIDIA GPU DRA driver (gpu.nvidia.com). 

## Testing

I tested the following scenarios on a two node k8s 1.34 cluster with two GPUs and the NVIDIA DRA driver installed. Steps to reproduce:

1. Create a pod that uses a GPU via a directly-named DRA ResourceClaim (allocated by gpu.nvidia.com) alongside a non-GPU pod on the same node. Verify the claim is allocated and reserved for the pod.
2. Trigger the driver-manager eviction flow (uninstall_driver). Verify the DRA GPU pod is identified and evicted, while the non-GPU pod is skipped.
3. Verify traditional device-plugin pods are still detected via nvidia.com/gpu and nvidia.com/mig-* resource requests.

### DIFF
--- a/internal/kubernetes/client.go
+++ b/internal/kubernetes/client.go
@@ -26,6 +26,8 @@ import (
 
 	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
+	resourcev1 "k8s.io/api/resource/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/labels"
@@ -42,6 +44,7 @@ const (
 	nvidiaDomainPrefix       = "nvidia.com"
 	nvidiaResourceNamePrefix = nvidiaDomainPrefix + "/" + "gpu"
 	nvidiaMigResourcePrefix  = nvidiaDomainPrefix + "/" + "mig-"
+	nvidiaDRADriverName      = "gpu." + nvidiaDomainPrefix
 
 	kubeClientPollInterval = 5 * time.Second
 )
@@ -239,8 +242,11 @@ func (c *Client) DeleteOrEvictPods(nodeName string, drainOpts DrainOptions) erro
 	c.log.Infof("Draining node %s of any GPU pods", nodeName)
 
 	customDrainFilter := func(pod corev1.Pod) drain.PodDeleteStatus {
-		deletePod := gpuPodSpecFilter(pod)
-		if !deletePod {
+		usesGPU, err := c.podUsesGPU(pod)
+		if err != nil {
+			return drain.MakePodDeleteStatusWithError(err.Error())
+		}
+		if !usesGPU {
 			return drain.MakePodDeleteStatusSkip()
 		}
 		return drain.MakePodDeleteStatusOkay()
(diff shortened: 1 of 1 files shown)


shivamerla (approved):
Looks good!

internal/kubernetes/client.go
```diff
@@ -288,6 +301,58 @@ func gpuPodSpecFilter(pod corev1.Pod) bool {
...
+			return *status.ResourceClaimName
+		}
+	}
+	return ""
+}
+
+// isAllocatedNvidiaGPUClaim reports whether a ResourceClaim has been allocated a
+// device by the NVIDIA GPU DRA driver.
+func isAllocatedNvidiaGPUClaim(claim *resourcev1.ResourceClaim) bool {
+	if claim.Status.Allocation == nil {
```
tariq1890, line 383: do we need to nil check `claim.Status` first?

internal/kubernetes/client.go
```diff
@@ -288,6 +301,58 @@ func gpuPodSpecFilter(pod corev1.Pod) bool {
...
+		}
+	}
+
+	return c.podUsesGPUViaDRA(pod)
+}
+
+// podUsesGPUViaDRA reports whether the pod holds a ResourceClaim allocated by
+// the NVIDIA GPU DRA driver. It resolves the name of each claim the pod
+// references and inspects that claim's allocation.
+func (c *Client) podUsesGPUViaDRA(pod corev1.Pod) (bool, error) {
```
tariq1890:
```suggestion
func (c *Client) podHasGPUResourceClaim(pod corev1.Pod) (bool, error) {
```

internal/kubernetes/client.go
karthikvetrivel: Good point, I don't think we need to since we start by looping through the reference claims in a given pod. Will remove. 

EDIT: readded.

internal/kubernetes/client.go
karthikvetrivel: Resolved!

internal/kubernetes/client.go
```diff
@@ -288,6 +301,58 @@ func gpuPodSpecFilter(pod corev1.Pod) bool {
...
+}
+
+// isAllocatedNvidiaGPUClaim reports whether a ResourceClaim has been allocated a
+// device by the NVIDIA GPU DRA driver.
+func isAllocatedNvidiaGPUClaim(claim *resourcev1.ResourceClaim) bool {
+	if claim.Status.Allocation == nil {
+		return false
+	}
+	for _, result := range claim.Status.Allocation.Devices.Results {
+		if result.Driver == nvidiaDRADriverName {
```
tariq1890, line 387:
```suggestion
	if claim.Status != nil && claim.Status.Allocation != nil && claim.Status.Allocation.Devices != nil {
	    for _, result := range claim.Status.Allocation.Devices.Results {
		    if result.Driver == nvidiaDRADriverName {
```

karthikvetrivel: I'm not sure this will compile (verified). I re-added the Allocation check though.

tariq1890 (approved):

tariq1890 (approved):

PR #205, opened 2026-07-06
add ability to bounce custom gpu client pods

tariq1890:
This PR is linked to https://github.com/NVIDIA/gpu-operator/pull/2607.

This adds the ability for the `k8s-driver-manager` to detect third-party/custom gpu client pods during driver upgrades and to evict them as they too will be holding onto gpu device handles. With this change, users can deploy their custom gpu client pods and also take advantage of the automatic toggling of gpu client components (just as what is done for the first-class gpu clients namely the "operands") during a driver upgrade.

**NOTE**: The third-party gpu client pods MUST add the `nvidia.com/gpu.deploy.client: true` label to its nodeSelector so that the k8s-driver-manager detects and performs the automated toggle during a driver upgrade

### DIFF
--- a/internal/kubernetes/client.go
+++ b/internal/kubernetes/client.go
@@ -27,6 +27,7 @@ import (
 	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
@@ -41,6 +42,8 @@ const (
 	nvidiaDomainPrefix       = "nvidia.com"
 	nvidiaResourceNamePrefix = nvidiaDomainPrefix + "/" + "gpu"
 	nvidiaMigResourcePrefix  = nvidiaDomainPrefix + "/" + "mig-"
+
+	kubeClientPollInterval = 5 * time.Second
 )
 
 // Client represents a Kubernetes client wrapper use to perform all the Kubernetes operations required by k8s-driver-manager
@@ -169,10 +172,10 @@ func (c *Client) UncordonNode(nodeName string) error {
 func (c *Client) WaitForPodTermination(selectorMap map[string]string, namespace, nodeName string, timeout time.Duration) error {
 	selector := labels.SelectorFromSet(selectorMap)
 
-	return wait.PollUntilContextTimeout(c.ctx, 5*time.Second, timeout, true, func(ctx context.Context) (bool, error) {
+	return wait.PollUntilContextTimeout(c.ctx, kubeClientPollInterval, timeout, true, func(ctx context.Context) (bool, error) {
 		pods, err := c.clientset.CoreV1().Pods(namespace).List(c.ctx, metav1.ListOptions{
 			LabelSelector: selector.String(),
-			FieldSelector: "spec.nodeName=" + nodeName,
+			FieldSelector: fields.OneTermEqualSelector("spec.nodeName", nodeName).String(),
 		})
 		if err != nil {
 			return false, err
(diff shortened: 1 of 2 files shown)


internal/kubernetes/client.go
```diff
@@ -183,6 +184,68 @@ func (c *Client) WaitForPodTermination(selectorMap map[string]string, namespace,
...
+// has the specified key in its nodeSelector to terminate.
+//
+// Owner controller lookups are cached by UID across poll iterations. We don't expect the owners to actually change
+// during the lifetime of the wait, so caching the owners list optimises for kube API call volumes.
+func (c *Client) WaitForPodsWithParentNodeSelector(nodeName, nodeSelectorKey string, timeout time.Duration) error {
+	// ownerMatchCache maps owner UID → whether its nodeSelector contains nodeSelectorKey.
+	// Owner controllers are stable during the wait; only the pods themselves are terminating.
+	ownerMatchCache := make(map[types.UID]bool)
+
+	return wait.PollUntilContextTimeout(c.ctx, 5*time.Second, timeout, true, func(ctx context.Context) (bool, error) {
```
rahulait: Can we move `5*time.Second` to a constant or variable and refer it here?

internal/kubernetes/client.go
```diff
@@ -183,6 +184,68 @@ func (c *Client) WaitForPodTermination(selectorMap map[string]string, namespace,
...
+		podList, err := c.clientset.CoreV1().Pods(corev1.NamespaceAll).List(ctx, metav1.ListOptions{
+			FieldSelector: "spec.nodeName=" + nodeName,
```
rahulait:
```suggestion
			FieldSelector: fields.OneTermEqualSelector("spec.nodeName", nodeName).String(),
```

internal/kubernetes/client.go
```diff
@@ -183,6 +184,68 @@ func (c *Client) WaitForPodTermination(selectorMap map[string]string, namespace,
...
+		})
+		if err != nil {
+			return false, fmt.Errorf("failed to list pods on node %s: %w", nodeName, err)
+		}
+
+		matchCount := 0
+		for _, pod := range podList.Items {
+			for _, ownerRef := range pod.OwnerReferences {
```
rahulait: Instead of iterating over all ownerReferences, can we just use the single owner reference which has controller: true, by using:
```
controllerRef := metav1.GetControllerOf(&pod)
if controllerRef == nil || controllerRef.Kind != "DaemonSet" {
	continue
}
```

internal/kubernetes/client.go
```diff
@@ -183,6 +184,68 @@ func (c *Client) WaitForPodTermination(selectorMap map[string]string, namespace,
...
+				matches, ok := ownerMatchCache[ownerRef.UID]
+				if !ok {
+					nodeSelector, err := c.getOwnerNodeSelector(ctx, pod.Namespace, ownerRef)
+					if err != nil {
+						c.log.Warnf("transient error checking parent nodeSelector for pod %s/%s, will retry: %v", pod.Namespace, pod.Name, err)
+						return false, nil
+					}
+					_, matches = nodeSelector[nodeSelectorKey]
+					ownerMatchCache[ownerRef.UID] = matches
```
rahulait: Should we ignore labels like `nvidia.com/gpu.deploy.client: randomvalue`?

internal/kubernetes/client.go
tariq1890: Thanks

internal/kubernetes/client.go
tariq1890: Good point, I guess we can

tariq1890: I am no longer doing KubeClient get call to the Daemon Set, so I don't need the cache anymore

internal/kubernetes/client.go
tariq1890: This is done

internal/kubernetes/client.go
tariq1890: Done

internal/kubernetes/client.go
tariq1890: Done

internal/kubernetes/client.go
```diff
@@ -183,6 +186,35 @@ func (c *Client) WaitForPodTermination(selectorMap map[string]string, namespace,
...
+			return false, fmt.Errorf("failed to list pods on node %s: %w", nodeName, err)
+		}
+
+		matchCount := 0
+		for _, pod := range podList.Items {
+			ownerRef := metav1.GetControllerOf(&pod)
+			if ownerRef == nil || ownerRef.Kind != "DaemonSet" {
+				continue
+			}
+			if _, ok := pod.Spec.NodeSelector[nodeSelectorKey]; ok {
```
rahulait, line 206: We still should look at the value. Consider this case where a user had that label as true initially, and later they decided to just set it to false instead of deleting it. So like:
```
nvidia.com/gpu.deploy.client: false
```

In this case, we'll still be waiting for pod termination and iterating with matchCount > 0 until timeout happens

rahulait: Since the value can be "true" or "paused-for-driver-upgrade", we should handle this cleanly if someone sets it to some other value.

rahulait (approved):
LGTM

internal/kubernetes/client.go
tariq1890: As discussed offline, this is a known issue with the "deploy label" approach where any manual value set to it by a user may not be honoured by the operator and will eventually be overridden by the gpu-operator. This limitation is pre-existing.

PR #210, opened 2026-07-13
retry the cordon and uncordon of a node on transient API errors

### LINKED ISSUE #190: [Bug]: k8s-driver-manager exits 0 when post-driver uncordon fails, leaving node Ready,SchedulingDisabled
**Describe the bug**

`k8s-driver-manager` (run as the init container of `nvidia-driver-daemonset`) cordons the node, performs driver work, then calls `kubectl uncordon` and exits. If that final `kubectl uncordon` call fails for a transient reason (API-server timeout, 5xx, connection refused, brief auth blip), the script does not propagate the error — the init container exits 0 and the pod becomes Ready.

The Node itself stays `Ready,SchedulingDisabled`. From the DaemonSet's perspective everything looks healthy, so no controller retries; from the scheduler's perspective the node is silently removed from the schedulable GPU pool. The failure mode is invisible until something downstream notices missing capacity.

**To Reproduce**

1. Deploy the GPU Operator (v25.10.1 in our case) on a cluster where the driver daemonset uses `k8s-driver-manager` as an init container.
2. Trigger a driver-manager run on a GPU node (e.g. a driver upgrade, or a manual restart of the driver pod).
3. While the init container is past the driver-work phase and about to run `kubectl uncordon`, induce a transient API-server failure for that call — easiest reliable reproducer is to interpose a deny / 503 on the `PATCH .../nodes/<name>` request, or to network-partition the kubelet briefly so the in-pod kubectl call times out.
4. Observe:
   - `k8s-driver-manager` init container exits with status 0.
   - `nvidia-driver-daemonset` pod is Ready.
   - `kubectl get node <name>` shows `Ready,SchedulingDisabled` and remains in that state indefinitely; no further reconciliation occurs.

**Expected behavior**

`k8s-driver-manager` should not silently succeed when the final uncordon fails. Specifically:

1. Retry the uncordon a bounded number of times with exponential backoff on transient errors (timeouts, 5xx, connection refused). The retry budget should be configurable via env var.
2. If retries are exhausted, exit non-zero. The init container should fail loudly, the daemonset pod should report not-ready, and the next driver-manager run should retry from a known state rather than leaving the node stuck cordoned.

 - `k8s-driver-manager` source (Commit SHA or image digest): `nvcr.io/nvidia/cloud-native/k8s-driver-manager:v0.9.1` @ `sha256:c549346eb993fda62e9bf665aabaacc88abc06b0b24e69635427d4d71c2d5ed4`
 - Host OS: Ubuntu 22.04.5 LTS
 - Kernel Version: 5.15.0-173-generic
 - NVIDIA Driver Version: 580.105.08
 - Container Runtime Version: containerd 2.1.1
 - CPU Architecture: x86_64

 If applicable, also provide:
 - Kubernetes Distro and Version: upstream Kubernetes v1.35.3 (CAPI-managed clusters)
 - NVIDIA GPU Operator version: v25.10.1

**Information to attach** (optional if deemed irrelevant)

- `k8s-driver-manager` init container logs from an affected pod (can attach a redacted sample on request).
- `kubectl describe node` for a node stuck in `Ready,SchedulingDisabled` after a driver-manager run.

---

We are mitigating downstream with a Kubernetes controller that observes the stuck state and re-uncordons the node, but the root-cause fix belongs in `k8s-driver-manager` so other operators don't have to carry the safety net. Happy to send a PR if the maintainers agree on the approach (retry + non-zero exit on terminal failure).

AcidLeroy:
Relates to #190 — see the scope note at the end for why this does not close it.

`k8s-driver-manager` cordons the node, performs its driver work, then uncordons it. Both calls gave up on the first API error, so a transient API-server timeout, 5xx, or connection refused could abort an upgrade or leave the node cordoned. We hit this as our clusters come up for the first time, when the API server is not yet fully ready; it is not an issue during steady state or upgrades.

This wraps both calls in a bounded exponential backoff (5 attempts, 1s base, factor 2, 20% jitter — roughly 15s of retrying):

- The retry policy lives at the call site, in driver-manager's `cordonNode` and `uncordonNode`, so the kube client stays a thin wrapper over the Kubernetes API and carries no retry concern. `internal/kubernetes/client.go`, `go.mod` and `vendor/` are unchanged by this PR.
- Because the client reads the Node inside `CordonNode`/`UncordonNode`, a retried call always patches a freshly read object rather than a stale one.
- How each call site treats an exhausted budget is unchanged from today's behaviour.
- The retry loop is context aware: a cancelled context ends it promptly instead of sleeping out the remaining backoff, and is joined with the last error via `errors.Join` so both causes stay inspectable.
- Failed attempts are logged with their attempt number and only described as retrying while attempts remain, so the attempt that exhausts the budget is logged as a final failure instead.

There is no new env var or flag — the retry limit is a constant.

Tests: the retry helper is unit tested directly. The client's per-attempt contract is covered by driving the real typed clientset against an `httptest` stand-in API server with independent failure budgets for reads and writes, so both a failed GET and a failed PATCH are exercised. (`k8s.io/client-go/kubernetes/fake` is not vendored, and vendoring it would add a large unrelated diff.)

**Scope note.** Earlier revisions of this PR also propagated a failed uncordon so the init container exits non-zero instead of logging a warning and exiting 0. Per review that behavioural change is dropped here and will follow separately, leaving this PR limited to the retries. It is also why this does not close #190: the retries make a stuck `Ready,SchedulingDisabled` node far less likely, but an exhausted budget still leaves it cordoned silently.

### DIFF
--- a/cmd/driver-manager/main.go
+++ b/cmd/driver-manager/main.go
@@ -40,13 +40,14 @@ import (
 )
 
 const (
-	driverRoot            = "/run/nvidia/driver"
-	driverPIDFile         = "/run/nvidia/nvidia-driver.pid"
-	driverConfigStateFile = "/run/nvidia/nvidia-driver.state"
-	operatorNamespace     = "gpu-operator"
-	pausedStr             = "paused-for-driver-upgrade"
-	defaultDrainTimeout   = time.Second * 0
-	defaultGracePeriod    = 5 * time.Minute
+	driverRoot             = "/run/nvidia/driver"
+	driverPIDFile          = "/run/nvidia/nvidia-driver.pid"
+	driverConfigStateFile  = "/run/nvidia/nvidia-driver.state"
+	operatorNamespace      = "gpu-operator"
+	pausedStr              = "paused-for-driver-upgrade"
+	defaultDrainTimeout    = time.Second * 0
+	defaultGracePeriod     = 5 * time.Minute
+	defaultUncordonRetries = 5
 
 	nvidiaDomainPrefix = "nvidia.com"
 
@@ -402,8 +403,8 @@ func (dm *DriverManager) uninstallDriver() error {
 
 	// Cleanup and reschedule components
 	if dm.isGPUPodEvictionEnabled() || dm.isAutoDrainEnabled() {
-		if err := dm.kubeClient.UncordonNode(dm.config.nodeName); err != nil {
-			dm.log.Warnf("Failed to uncordon node: %v", err)
+		if err := dm.kubeClient.UncordonNode(dm.config.nodeName, kube.RetryBackoff(defaultUncordonRetries)); err != nil {
+			return fmt.Errorf("failed to uncordon node: %w", err)
 		}
 	}
 
@@ -980,7 +981,7 @@ func (dm *DriverManager) cleanupOnFailure() {
 
 	switch {
 	case managerCanEvict:
-		if err := dm.kubeClient.UncordonNode(dm.config.nodeName); err != nil {
+		if err := dm.kubeClient.UncordonNode(dm.config.nodeName, kube.RetryBackoff(defaultUncordonRetries)); err != nil {
 			dm.log.Warnf("Failed to uncordon node during cleanup: %v", err)
 		}
 	case policyEnabled:
--- a/cmd/driver-manager/main.go
+++ b/cmd/driver-manager/main.go
@@ -47,6 +47,10 @@ const (
 	pausedStr             = "paused-for-driver-upgrade"
 	defaultDrainTimeout   = time.Second * 0
 	defaultGracePeriod    = 5 * time.Minute
+	// defaultCordonRetries bounds the retries for the cordon and uncordon calls
+	// that bracket the driver uninstall, so a transient API server error does not
+	// abort an upgrade or leave the node cordoned.
+	defaultCordonRetries = 5
 
 	nvidiaDomainPrefix = "nvidia.com"
 
@@ -344,7 +348,7 @@ func (dm *DriverManager) uninstallDriver() error {
 		}
 
 		if dm.isGPUPodEvictionEnabled() || dm.isAutoDrainEnabled() {
-			if err := dm.kubeClient.UncordonNode(dm.config.nodeName); err != nil {
+			if err := dm.uncordonNode(); err != nil {
 				dm.log.Warnf("Failed to uncordon node: %v", err)
 			}
 		}
@@ -407,7 +411,7 @@ func (dm *DriverManager) uninstallDriver() error {
 	// the post-unload auto-drain fallback runs after the kubelet-plugin is gone,
 	// leaving drained claim-holders stuck in Terminating.
 	if dm.isGPUPodEvictionEnabled() || (dm.components.draDriverDeployed != "" && dm.isAutoDrainEnabled()) {
-		if err := dm.kubeClient.CordonNode(dm.config.nodeName); err != nil {
+		if err := dm.cordonNode(); err != nil {
 			return fmt.Errorf("failed to cordon node: %w", err)
 		}
 
@@ -500,7 +504,7 @@ func (dm *DriverManager) uninstallDriver() error {
 
 	// Cleanup and reschedule components
 	if dm.isGPUPodEvictionEnabled() || dm.isAutoDrainEnabled() {
-		if err := dm.kubeClient.UncordonNode(dm.config.nodeName); err != nil {
+		if err := dm.uncordonNode(); err != nil {
 			dm.log.Warnf("Failed to uncordon node: %v", err)
 		}
 	}
@@ -1123,6 +1127,21 @@ func (dm *DriverManager) isDriverAutoUpgradePolicyEnabled() bool {
 	return false
 }
 
+// cordonNode and uncordonNode wrap the kube client calls with this command's
+// retry policy. The policy lives here rather than in the client so the client
+// stays a thin wrapper over the Kubernetes API.
+func (dm *DriverManager) cordonNode() error {
+	return retryOnAnyError(dm.ctx, dm.log, retryBackoff(defaultCordonRetries),
+		fmt.Sprintf("cordon node %s", dm.config.nodeName),
+		func() error { return dm.kubeClient.CordonNode(dm.config.nodeName) })
+}
+
+func (dm *DriverManager) uncordonNode() error {
+	return retryOnAnyError(dm.ctx, dm.log, retryBackoff(defaultCordonRetries),
+		fmt.Sprintf("uncordon node %s", dm.config.nodeName),
+		func() error { return dm.kubeClient.UncordonNode(dm.config.nodeName) })
+}
+
 func (dm *DriverManager) cleanupOnFailure() {
 	dm.log.Info("Performing cleanup on failure")
 	policyEnabled := dm.isDriverAutoUpgradePolicyEnabled()
@@ -1130,7 +1149,7 @@ func (dm *DriverManager) cleanupOnFailure() {
 
 	switch {
 	case managerCanEvict:
-		if err := dm.kubeClient.UncordonNode(dm.config.nodeName); err != nil {
+		if err := dm.uncordonNode(); err != nil {
 			dm.log.Warnf("Failed to uncordon node during cleanup: %v", err)
 		}
 	case policyEnabled:
--- a/internal/kubernetes/client.go
+++ b/internal/kubernetes/client.go
@@ -111,16 +111,7 @@ func (c *Client) UpdateNodeLabels(nodeName string, nodeLabels map[string]string)
 		return fmt.Errorf("failed to marshal patch: %w", err)
 	}
 
-	backoff := wait.Backoff{
-		Duration: time.Second,
-		Factor:   2.0,
-		Jitter:   0.2,
-		Steps:    7,
-	}
-
-	return retry.OnError(backoff, func(err error) bool {
-		return true
-	}, func() error {
+	return retryOnAnyError(RetryBackoff(7), func() error {
 		_, err := c.clientset.CoreV1().Nodes().Patch(c.ctx, nodeName, types.StrategicMergePatchType, patchBytes, metav1.PatchOptions{})
 		if err != nil {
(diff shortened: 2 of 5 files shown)


AcidLeroy:
@tariq1890 Do I need to mention this in an office hours or something? Happy to discuss the direction/motivation here.

cmd/driver-manager/main.go
```diff
@@ -219,6 +221,13 @@ func main() {
 			EnvVars:     []string{"FORCE_REINSTALL"},
 			Value:       false,
 		},
+		&cli.IntFlag{
+			Name:        "uncordon-retries",
```
tariq1890: I'd prefer not exposing a knob for this and go with a default retry limit for now

internal/kubernetes/client_test.go
```diff
@@ -0,0 +1,59 @@
+/*
+ * Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
```
tariq1890: Year needs fixing

tariq1890:
Thank you for your PR. We are in somewhat of  a code freeze as we have a new planned release coming up. I will need to discuss with the rest of the team if we are to include this change in the next release.

AcidLeroy:
> Thank you for your PR. We are in somewhat of a code freeze as we have a new planned release coming up. I will need to discuss with the rest of the team if we are to include this change in the next release.

Thanks @tariq1890! This was just a weirdness that we noticed as our clusters come up for the first time (api server might not be fully ready). It's not too much of an issue during steady state or upgrades. Happy to discuss other ways to handle this as well if this doesn't work for you.

cmd/driver-manager/main.go
AcidLeroy: Removed the knob — dropped the `uncordon-retries` flag, `UNCORDON_RETRIES` env var, and config field. Now uses the `defaultUncordonRetries` (5) constant directly for both uncordon calls.

internal/kubernetes/client_test.go
AcidLeroy: Fixed — bumped to 2026.

AcidLeroy:
@tariq1890 any updates on getting this in?

tariq1890:
@AcidLeroy Apologies for the delayed responses. We are in a code-freeze right now as we are working on getting the next release of the gpu-operator out. Once that's done, we can get back to reviewing this PR. Should this PR be merged, it will be in the releases of gpu-operator following the next.

kvalliyurnatt (dismissed):

kvalliyurnatt:
@AcidLeroy regarding
>  the symmetric CordonNode still aborts on the first transient error (unchanged here) — only a failed uncordon leaves the node stuck Ready,SchedulingDisabled. Happy to extend the same retry to cordon in a follow-up if maintainers prefer.

could you include this in the existing PR, I think that would be a good to have as well.

Also could you update the PR description, it seems outdated

AcidLeroy:
> @AcidLeroy regarding
> 
> > the symmetric CordonNode still aborts on the first transient error (unchanged here) — only a failed uncordon leaves the node stuck Ready,SchedulingDisabled. Happy to extend the same retry to cordon in a follow-up if maintainers prefer.
> 
> could you include this in the existing PR, I think that would be a good to have as well.
> 
> Also could you update the PR description, it seems outdated

@kvalliyurnatt 
Updated the description and also included retries for cordon to keep it symmetric.

internal/kubernetes/client.go
```diff
@@ -144,30 +135,61 @@ func (c *Client) GetNodeAnnotationValue(nodeName, annotation string) (string, er
...
+// Kubernetes API errors, bounded to the given number of attempts (at least one).
+func RetryBackoff(attempts int) wait.Backoff {
+	if attempts < 1 {
+		attempts = 1
+	}
+	return wait.Backoff{Duration: time.Second, Factor: 2.0, Jitter: 0.2, Steps: attempts}
+}
+
+// retryOnAnyError retries fn on any error using the given backoff.
+func retryOnAnyError(backoff wait.Backoff, fn func() error) error {
```
kvalliyurnatt: nit: could we make this function context aware, so that when context is cancelled we return promptly instead of sleeping through remaining backoff

internal/kubernetes/client.go
```diff
@@ -144,30 +135,61 @@ func (c *Client) GetNodeAnnotationValue(nodeName, annotation string) (string, er
...
+	if cordon {
+		verb = "cordon"
 	}
 
-	drainHelper := &drain.Helper{Ctx: c.ctx, Client: c.clientset}
-	return drain.RunCordonOrUncordon(drainHelper, node, false)
+	return retryOnAnyError(backoff, func() error {
+		node, err := c.clientset.CoreV1().Nodes().Get(c.ctx, nodeName, metav1.GetOptions{})
+		if err != nil {
+			c.log.Warnf("Failed to get node %s to %s it, retrying: %v", nodeName, verb, err)
```
kvalliyurnatt: nit: This reports retrying even on the last failure, can we maybe log the retry count and log the last attempt.

internal/kubernetes/client_test.go
```diff
@@ -0,0 +1,190 @@
...
+	require.NoError(t, err)
+
+	log := logrus.New()
+	log.SetOutput(io.Discard)
+
+	return &Client{ctx: context.Background(), log: log, clientset: clientset}
+}
+
+func TestCordonNodeRetriesUntilSuccess(t *testing.T) {
+	api := &flakyNodeAPI{nodeName: "gpu-node", failuresLeft: 2}
```
kvalliyurnatt: nit: The current failures are consumed by the initial GET requests, so these tests do not exercise a failed PATCH . Could we add PATCH-specific failure coverage ?

kvalliyurnatt (approved):
LGTM, left a few nits, but non-blocking.

internal/kubernetes/client.go
```diff
@@ -144,30 +140,99 @@ func (c *Client) GetNodeAnnotationValue(nodeName, annotation string) (string, er
...
+		lastErr = fn()
+		return lastErr == nil, nil
+	})
+	if err == nil {
+		return nil
+	}
+
+	if ctxErr := ctx.Err(); ctxErr != nil {
+		if lastErr != nil {
+			return fmt.Errorf("%w (last error: %v)", ctxErr, lastErr)
```
tariq1890: why not use errors.Join instead?

internal/kubernetes/client.go
```diff
@@ -144,30 +140,99 @@ func (c *Client) GetNodeAnnotationValue(nodeName, annotation string) (string, er
 	return node.Annotations[annotation], nil
 }
 
-// CordonNode cordons a Node given a Node name marking it as Unschedulable
-func (c *Client) CordonNode(nodeName string) error {
+// CordonNode cordons a Node given a Node name marking it as Unschedulable.
+// It retries on any error until the backoff budget is exhausted, then returns
+// the last error.
+func (c *Client) CordonNode(nodeName string, backoff wait.Backoff) error {
```
tariq1890: I don't think we should be adding the backoff to the function signature of the kube client.

We could do either of the two:

- Add the backoff settings to the Client struct
- Implement the retry logic at the call site layer instead.

I prefer the second option.

AcidLeroy:
@tariq1890 addressed comments. Please take another look when you have time.

internal/kubernetes/client.go
```diff
@@ -144,30 +132,89 @@ func (c *Client) GetNodeAnnotationValue(nodeName, annotation string) (string, er
...
+	}
+	return wait.Backoff{Duration: time.Second, Factor: 2.0, Jitter: 0.2, Steps: attempts}
+}
+
+// RetryOnAnyError retries fn on any error using the given backoff, logging each
+// failed attempt at action. A cancelled ctx ends the retry loop promptly
+// instead of sleeping out the remaining backoff, and is joined with the last
+// error rather than being reduced to it, which wait.Interrupted would
+// otherwise conflate.
+func RetryOnAnyError(ctx context.Context, log *logrus.Logger, backoff wait.Backoff, action string, fn func() error) error {
```
tariq1890: let's not define the method here. It is better to declare it in the `main` package.

internal/kubernetes/client.go
```diff
@@ -114,20 +114,8 @@ func (c *Client) UpdateNodeLabels(nodeName string, nodeLabels map[string]string)
...
-		Duration: time.Second,
-		Factor:   2.0,
-		Jitter:   0.2,
-		Steps:    7,
-	}
-
-	return retry.OnError(backoff, func(err error) bool {
-		return true
-	}, func() error {
+	return RetryOnAnyError(c.ctx, c.log, RetryBackoff(7), fmt.Sprintf("update labels on node %s", nodeName), func() error {
```
tariq1890: Similar to my previous comment, it's probably better to move the retry logic one layer up. I suggest doing either of the two:

- revert this change to what it was before
- move the retry block to where the `UpdateNodeLabels` is invoked.

tariq1890:
@AcidLeroy I was thinking we limit the scope of this PR to adding retries for the kube cordon and uncordon actions. The failing fast on uncordon fail does make sense, but it is a behavioural change to an execution flow which hasn't been changed in a long while. A retry logic would have averted the original issue you when your env encountered the Uncordon failure, yes? I am just being cautious here.

If we feel strongly about introducing this new behaviour of exiting with a non-zero code, we could perhaps explore that in a separate PR/commit? Let me know what you think.

AcidLeroy:
@tariq1890 thanks again for the review. 

> A retry logic would have averted the original issue you when your env encountered the Uncordon failure, yes? I am just being cautious here.

Yes, eventually everything came up healthy so having a retry, in theory, should the original bug I had. 


> If we feel strongly about introducing this new behaviour of exiting with a non-zero code, we could perhaps explore that in a separate PR/commit?

That sounds like a wise decision, agreed.

tariq1890:
Thank you for your patience and diligence with this @AcidLeroy, it is much appreciated!

The changes, the updated PR titles and commit message all LGTM. The only thing remaining is the signing the commit. See [here](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits).

tariq1890:
It needs to show up as `Verified`

Screenshot 2026-09-08 at 7 50 36 PM

AcidLeroy:
> It needs to show up as `Verified`

Should be fixed now.

tariq1890 (approved):

PR #217, opened 2026-07-28
Skip eviction when no driver is loaded

### LINKED ISSUE #213: [Bug]: Deadlock with DRA on node reboot
**Describe the bug**
There is a circular dependency when using DRA on node reboots when running uninstall in the GPU driver init containers:
- the DRA plugin requires GPU drivers (and/or the toolkit, which depends on it):
```
[dra plugin logs]
2026-07-20T10:42:10Z  /driver-root (/run/nvidia/driver on host): nvidia-smi: not found, libnvidia-ml.so.1: not found, current contents: [lib].
```
- during a reboot, the GPU driver container restarts and its k8s-driver-manager init container tries the uninstall that deletes all GPU consumers
```
[gpu-driver-container logs]
time=2026-07-20T10:41:53Z level=info msg=Identifying GPU pods to delete
```
- to delete a GPU consumer that uses DRA, it needs the DRA plugin to be registered (CDI/MIG cleanup); if the latter is not running, the pod stays in `Terminating` forever
```
[pod events]
  Warning  FailedPrepareDynamicResources  26s (x18 over 21m)  kubelet             Failed to prepare dynamic resources: prepare dynamic resources: DRA driver gpu.nvidia.com is not registered
```

Note that this is not a DRA issue, on kubelet side this behaviour is intentional: https://github.com/kubernetes/kubernetes/issues/129402#issuecomment-2577794262

**To Reproduce**
- run any workload that request a GPU on a DRA-enabled GPU infrastructure
- reboot the associated node
- check that both gpu-driver pod is pending in its k8s-driver-manager phase

**Expected behavior**
The k8s-driver-manager should support this use case and not cause a deadlock

 - `k8s-driver-manager` source (Commit SHA or image digest): main
 - Host OS: all
 - Kernel Version: all
 - NVIDIA Driver Version: all
 - Container Runtime Version: all
 - CPU Architecture: all
 - GPU Model(s): all

karthikvetrivel:
> **Stacked on #203.** The base branch `kv-drain-dra-operands` mirrors #203's tip, so the diff here is only this PR's commit. After #203 merges I will retarget this PR to `main` and rebase.

Fixes #213.

## Problem

On a node reboot or fresh install, the driver-manager init container evicts all GPU operator components and drains GPU workloads before it allows the driver install to proceed. On a DRA node this deadlocks. A claim-holding pod cannot finish terminating until the kubelet calls `NodeUnprepareResources` on the DRA kubelet-plugin, and this obligation survives the reboot because the kubelet checkpoints prepared claims in `/var/lib/kubelet/dra_manager_state` (this is intentional kubelet behavior, see kubernetes/kubernetes#129402). The plugin cannot start until the driver is installed, and the driver cannot install until driver-manager finishes its evictions. The evicted pods stay in `Terminating` forever, and the init container crashloops on the eviction timeout while flapping the `gpu.deploy.*` labels.

The eviction also protects nothing in this state. With no NVIDIA kernel modules loaded, no process on the node holds the driver.

## Fix

I added an early exit at the top of `uninstallDriver`, above `evictAllGPUOperatorComponents()`, which is where the deadlock happens. When no NVIDIA kernel modules are loaded, driver-manager skips eviction and performs only the steps that must precede a fresh install.

- It unmounts a stale `/run/nvidia/driver` rootfs and removes the stale PID file that a crashed driver container can leave behind.
- It unloads nouveau if present, since the installer relies on driver-manager for this.
- It restarts the DRA kubelet-plugin. I found that a plugin that ran against a previous driver container keeps its bind mount pinned to that rootfs and fails new `NodePrepareResources` calls against the replacement.
- It restores any `gpu.deploy.*` labels that a previous failed attempt left paused, then returns.

## Testing

I tested on a live single-GPU (Tesla T4) kubeadm cluster running the DRA stack, with a GPU workload pod holding an allocated ResourceClaim.

- I reproduced the deadlock with the base branch. After a reboot, the dra-validator was stuck in `Terminating`, the workload looped `FailedPrepareDynamicResources: DRA driver gpu.nvidia.com is not registered`, and the node stayed driver-less for 11 hours.
- I deployed this fix onto the deadlocked node and it recovered in about 5 minutes with no manual cleanup.
- I replaced the driver pod on a live node to confirm the kubelet-plugin restart is required. Without it, new claim prepares fail with `empty device edits`. With it, a fresh claim pod ran `nvidia-smi` about 20 seconds after the rotation.
- I rebooted the node with the fix in place. It recovered hands-off in about 7 minutes with zero stuck pods, and the workload kept its claim and GPU across the reboot.
- I verified the existing paths are unchanged. With the driver loaded, the same build ran the normal eviction and reload flow.

### DIFF
--- a/cmd/driver-manager/main.go
+++ b/cmd/driver-manager/main.go
@@ -296,6 +296,61 @@ func (dm *DriverManager) uninstallDriver() error {
 		return fmt.Errorf("failed to fetch auto upgrade annotation: %w", err)
 	}
 
+	// No driver loaded (node reboot or fresh install): nothing holds the driver, so
+	// skip eviction. Evicting here deadlocks on DRA nodes: terminating a claim-holding
+	// pod needs the kubelet-plugin, which cannot start until the driver install
+	// proceeds.
+	if !dm.isDriverLoaded() {
+		dm.log.Info("No NVIDIA driver loaded, skipping eviction of GPU clients and workloads")
+
+		// Restart the kubelet-plugin: it stays bound to the previous driver rootfs
+		// and would fail to prepare claims against the replacement. Drain it before
+		// unmounting that rootfs so the unmount is not done underneath it.
+		if err := dm.evictKubeletPlugin(); err != nil {
+			return fmt.Errorf("failed to evict DRA kubelet-plugin: %w", err)
+		}
+
+		// Clean up stale artifacts from a previous driver container
+		if err := dm.unmountRootfs(); err != nil {
+			return fmt.Errorf("failed to unmount stale rootfs: %w", err)
+		}
+		dm.removePIDFile()
+
+		// Unload nouveau if present; it blocks the driver install
+		if dm.isNouveauLoaded() {
+			if err := dm.unloadNouveau(); err != nil {
+				return fmt.Errorf("failed to unload nouveau driver: %w", err)
+			}
+			dm.log.Info("Successfully unloaded nouveau driver")
+		}
+
+		// Handle vfio-pci driver unbinding
+		if err := dm.unbindVfioPCI(); err != nil {
+			dm.log.Error("Unable to unbind vfio-pci driver from all devices")
+			return fmt.Errorf("failed to unbind vfio-pci driver: %w", err)
+		}
+
+		// Handle GPUDirect RDMA if enabled
+		// When GPUDirectRDMA is enabled, wait until MOFED driver has finished installing
+		if dm.isGPUDirectRDMAEnabled() {
+			dm.log.Info("GPUDirectRDMA is enabled, validating MOFED driver installation")
+			if err := dm.waitForMofedDriver(); err != nil {
+				return fmt.Errorf("failed to wait for MOFED driver: %w", err)
+			}
+		}
+
+		if dm.isGPUPodEvictionEnabled() || dm.isAutoDrainEnabled() {
+			if err := dm.kubeClient.UncordonNode(dm.config.nodeName); err != nil {
+				dm.log.Warnf("Failed to uncordon node: %v", err)
+			}
+		}
+
+		if err := dm.rescheduleGPUOperatorComponents(); err != nil {
+			return fmt.Errorf("failed to reschedule GPU operator components: %w", err)
+		}
+		return nil
+	}
+
 	// Always evict all GPU operator components across a driver restart. The DRA
 	// kubelet-plugin is the exception: it services NodeUnprepareResources for the
 	// claim-holders evicted here (e.g. dra-validator), so it must outlive them and


cmd/driver-manager/main.go
```diff
@@ -296,6 +297,40 @@ func (dm *DriverManager) uninstallDriver() error {
...
+			}
+			dm.log.Info("Successfully unloaded nouveau driver")
+		}
+
+		// Restart the kubelet-plugin: it stays bound to the previous driver rootfs
+		// and would fail to prepare claims against the replacement.
+		if err := dm.evictKubeletPlugin(); err != nil {
+			return fmt.Errorf("failed to evict DRA kubelet-plugin: %w", err)
+		}
+
```
rahulait, line 332: Lets also add here:
```
		if dm.isGPUPodEvictionEnabled() || dm.isAutoDrainEnabled() {
			if err := dm.kubeClient.UncordonNode(dm.config.nodeName); err != nil {
				dm.log.Warnf("Failed to uncordon node: %v", err)
			}
		}
```

Why this matters:

1. An upgrade attempt cordons a node.
2. The node reboots before that attempt reaches normal cleanup, or it reboots while the driver is absent.
3. node.spec.unschedulable: true survives the reboot.
4. On restart, this PR takes the new no-driver path, installs/re-enables operands, then returns without uncordoning.
5. DaemonSets recover because they tolerate being scheduled onto an unschedulable node, but ordinary workload Pods cannot be scheduled there.

So the driver may be healthy and DRA components may be running, yet the node stays permanently unavailable for newly scheduled GPU workloads.

cmd/driver-manager/main.go
```diff
@@ -296,6 +297,40 @@ func (dm *DriverManager) uninstallDriver() error {
 		return fmt.Errorf("failed to fetch auto upgrade annotation: %w", err)
 	}
 
+	// No driver loaded and no GPUs bound to vfio-pci (node reboot or fresh install):
+	// nothing holds the driver, so skip eviction. Evicting here deadlocks on DRA
+	// nodes: terminating a claim-holding pod needs the kubelet-plugin, which cannot
+	// start until the driver install proceeds. vfio-bound GPUs count as in use so
+	// that passthrough nodes keep the evict-then-unbind path.
+	if !dm.isDriverLoaded() && !dm.hasVFIOBoundGPUs() {
```
cdesiniotis: I am not following why we need to check whether GPUs are bound to `vfio-pci` module. The standard code path unconditionally unbinds all GPUs from the `vfio-pci` driver. Why don't we continue to do that in this case?
```suggestion
	if !dm.isDriverLoaded() {
```
Later in this if statement, we would make a call to `dm.unbindVfioPCI()`

cmd/driver-manager/main.go
cdesiniotis, line 332: The standard code path also has logic for wating on MOFED drivers. We should add that here as well:
```
	// Handle GPUDirect RDMA if enabled
	// When GPUDirectRDMA is enabled, wait until MOFED driver has finished installing
	if dm.isGPUDirectRDMAEnabled() {
		dm.log.Info("GPUDirectRDMA is enabled, validating MOFED driver installation")
		if err := dm.waitForMofedDriver(); err != nil {
			return fmt.Errorf("failed to wait for MOFED driver: %w", err)
		}
	}
```

cmd/driver-manager/main.go
```diff
@@ -296,6 +297,40 @@ func (dm *DriverManager) uninstallDriver() error {
...
+		dm.log.Info("No NVIDIA driver loaded, skipping eviction of GPU clients and workloads")
+
+		// Clean up stale artifacts from a previous driver container
+		if err := dm.unmountRootfs(); err != nil {
```
shivamerla, line 314: any specific reason the order is different here than https://github.com/karthikvetrivel/k8s-driver-manager/blob/4a612e3d17fffae7d2c4851ae358a78bea8497eb/cmd/driver-manager/main.go#L344?

i.e why don't we evict kubelet plugin, unmount and then reschedule again? That way unmount itself can happen cleanly.

cmd/driver-manager/main.go
karthikvetrivel: Here is what I was thinking: 

`isDriverLoaded()` only looks at `/sys/module/nvidia/refcnt`. On a node whose GPUs are bound to `vfio-pci`, the nvidia module isn't loaded, so `!isDriverLoaded()`is true not just after a reboot, but on every driver-manager run on that node. 

Without the second condition, a passthrough to driver-mode transition takes the early-exit path, which skips `evictAllGPUOperatorComponents()`, including sandbox-validator, sandbox-device-plugin, and vgpu-device-manager. We'd then unbind vfio-pci from devices that are still assigned.

cmd/driver-manager/main.go
```diff
@@ -758,6 +793,23 @@ func (dm *DriverManager) isDriverLoaded() bool {
...
 }
 
+// hasVFIOBoundGPUs reports whether any NVIDIA GPU on the node is bound to the
+// vfio-pci driver. On enumeration failure it returns true so the caller falls back
+// to evicting GPU clients as before.
+func (dm *DriverManager) hasVFIOBoundGPUs() bool {
+	gpus, err := nvpci.New().GetGPUs()
+	if err != nil {
+		dm.log.Warnf("Failed to enumerate NVIDIA PCI devices: %v", err)
+		return true
```
shivamerla: we should just return error here if PCIe enumeration fails. Let the caller handle these errors gracefully when needed.

cmd/driver-manager/main.go
```diff
@@ -758,6 +793,23 @@ func (dm *DriverManager) isDriverLoaded() bool {
...
+	}
+	for _, gpu := range gpus {
+		if gpu.Driver == "vfio-pci" {
```
shivamerla: On grace systems the driver name would be `nvgrace_vfio_pci`.

cmd/driver-manager/main.go
cdesiniotis: If we are transitioning the `gpu.workload.config` label from `vm-passthrough` to `container`, the operands that run for the `vm-passthrough` use case would automatically be evicted since the GPU Operator would remove their `deploy` labels. So I think it is fine to skip operand eviction (here in the k8s-driver-manager) in that case.

karthikvetrivel: Makes sense, removed.

cmd/driver-manager/main.go
karthikvetrivel: Remove since we no longer do a `vfio-pci` check.

cmd/driver-manager/main.go
karthikvetrivel: Removed since we no longer do a vfio-pci check.

cmd/driver-manager/main.go
karthikvetrivel: No particular reason. I re-ordered for consistency.

cmd/driver-manager/main.go
karthikvetrivel: Added.

karthikvetrivel: Added.

cdesiniotis (approved):

rahulait (approved):

rahulait:
I merged it and didn't realize it was not pointing to main. I created another PR which points to main and contains this change https://github.com/NVIDIA/k8s-driver-manager/pull/219

PR #219, opened 2026-07-29
Skip GPU client eviction when no NVIDIA driver is loaded

rahulait:
Pull request https://github.com/NVIDIA/k8s-driver-manager/pull/217 which was merged to different branch than main


cdesiniotis (approved):
Thanks for spotting this @rahulait

PR #227, opened 2026-08-07
Add THIRD_PARTY_NOTICES.md and a generator for it

abrarshivani:
Adds `THIRD_PARTY_NOTICES.md` for the Go deps linked into the two binaries this
repo ships, the script that generates it, and a CI job that fails when the
committed copy goes stale.

- `make third-party-notices` regenerates it
- `make check-third-party-notices` regenerates and diffs it, and runs in CI on every build
- Mirrors what was added to gpu-operator, adapted to this repo's build

## What to review

94% of the diff (8215 of 8734 added lines) is the generated
`THIRD_PARTY_NOTICES.md`. Another 58 are `go mod tidy` output. Hand-written is
461 lines, almost all one shell script:

| File | Added |
|---|---:|
| `scripts/generate-third-party-notices.sh` | 384 |
| `.github/workflows/third-party-notices-check.yaml` | 52 |
| `Makefile` | 18 |
| `.github/workflows/ci.yaml` | 4 |
| `.gitignore` | 2 |
| `deployments/devel/tools.go` | 1 |

`.gitignore` adds `/bin/` and `/.licenses-cache/`; `tools.go` pins
`go-licenses`. `deployments/devel/go.mod` and `go.sum` are `go mod tidy` output
from that one line, which moves `golang.org/x/tools` v0.31.0 to v0.36.0 in that
module. `golangci-lint`, `moq` and `go-licenses` were all built from it to
confirm nothing breaks.

## How it works

`make third-party-notices` runs `scripts/generate-third-party-notices.sh`:

1. Use `./bin/go-licenses` over anything on `PATH`, so the pinned version runs
2. Check the platform list against `DOCKER_BUILD_PLATFORM_OPTIONS` in
   `deployments/container/multi-arch.mk`, fail if out of sync
3. Run `go-licenses save` and `csv` over `./cmd/...` once per released platform
   (`GOFLAGS=-mod=vendor`, `CGO_ENABLED=0`), then merge. It resolves only the
   host platform and has no union mode.
4. Collapse to one row per package, joining every distinct license
5. Add `module@version` from `vendor/modules.txt` by longest-prefix match. In
   vendor mode `go-licenses` reports a URL into this repo at `HEAD`, which stops
   describing released content once `main` moves.
6. Write header, index table and license text to a temp file, then move it into
   place

## Implementation notes

- **Only the local module goes to `--ignore`.** It matches raw string prefixes,
  not path segments, so a short generic token silently drops whole ecosystems.
  `go-licenses` already omits the stdlib.
- **Licenses are joined, not picked.** `go-licenses` emits a row per recognized
  license and some entries here carry two, so key-only dedup would hide one.
- **The `Unknown` guard matches `Unknown` as a list element**, with
  `grep -qE '(^| / )Unknown( / |$)'`. Licenses are joined, so `grep -qx` misses
  `BSD-3-Clause / Unknown`. `go-licenses save` only warns on an unidentified
  license, so one would otherwise land in the file.
- **Output must match on every machine**, since CI diffs it. `LC_ALL=C` on every
  sort and grep. Whole-line `sort -u`, because key-only dedup keeps the first
  input row on BSD sort and applies a whole-line tiebreak on GNU sort. The awk
  counts instead of testing `pkg in lic`, which mawk and busybox awk get wrong
  (mawk is `/usr/bin/awk` on stock Debian). `grep -a` so a NUL byte cannot
  change the fence width.
- **Moved with `mv`, not `cp`.** `cp` truncates the target first, so an
  interrupted run would leave a half-written file.
- **`./bin/go-licenses` depends on `deployments/devel/go.{mod,sum}`**, so
  bumping the pin rebuilds it. Otherwise a stale local binary disagrees with
  CI's fresh build, showing up only as a stale-notices failure nobody can
  reproduce.
- **`check-third-party-notices` runs on every build**, not behind a changed-paths filter.
  The list is the import set of `./cmd/...`, so it goes stale when an ordinary
  `.go` file changes its imports, not only when `go.mod` or `vendor/` move.

## Scope

Covers `./cmd/...`, what `Dockerfile.distroless` builds and copies into the
image (`go build ./cmd/driver-manager/` and `go build ./cmd/vfio-manage/`).
Against `./...` the set is identical today, 503 packages either way, since
everything under `internal/` is reachable from `cmd/`. The narrower form keeps
tracking the image if an unshipped package is added later.

Not covered:

- The non-Go contents of the image: base image packages and the BusyBox tree,
  which have their own compliance processes.
- The shell scripts under `scripts/`, covered by this repo's `LICENSE`.
- The build toolchain in `deployments/devel/go.mod`. gpu-operator's toolchain
  module holds code generators whose output is committed and shipped; this one
  pins only `golangci-lint` and `moq`. The linter produces no artifact and there
  is no `go:generate` directive in the tree, so neither is part of the chain
  behind the released binaries. It could not be listed as things stand either:
  two of `golangci-lint`'s transitive deps ship a short-form Apache header the
  classifier cannot identify, and `go-licenses save` treats that as fatal.

## Shipping

Nothing ships the file today. It lives in the repo, and CI keeps it fresh.

- Not copied into the released image, so image size is unchanged. Same as
  gpu-operator.
- No release asset: this repo publishes no GitHub Releases, only tags, with
  releases going through the internal pipeline.

## Testing

- **Cross-host.** Same sha256
  `275452f423fa25608d4edd73512f18c1052967c29587ea8191333be19c613412` on macOS
  (BSD `awk` and `sort`) and in `golang:1.26.5` (`mawk`, GNU `sort`). Two runs
  on one host also match. Takes about 6 seconds once `go-licenses` is built.
- **Completeness** against `go list -deps ./cmd/...` for both platforms, minus
  stdlib and the local module: 497 third-party packages, 497 covered, 72 index
  entries, none unused, 0 missing, 0 uncovered. Five packages match two entries
  each: nested `third_party/forked/...` subtrees with their own license, where
  the more specific entry wins.
- **Content.** 72 index rows, 72 license-text sections, 0 `Unknown`, 0 "License
  text unavailable", 0 unresolved `module@version`. The multi-licensed entries
  (`sigs.k8s.io/json`, `sigs.k8s.io/yaml`) checked against `vendor/`.
- **The check passes and fails.** Exits 0 on the committed file. Exits 2 with
  `ERROR: THIRD_PARTY_NOTICES.md is stale` when it is modified, and
  `... is not tracked` when it is untracked, since `git diff` reports nothing
  for an untracked path.
- **Error paths.** Each fails with a useful message: platform matrix out of sync
  with `multi-arch.mk`, missing `vendor/modules.txt`, `modules.txt` unreadable
  as a module list, filesystem `replace` with no version. A versioned `replace`
  is attributed to the replacement module, which is what is vendored.
- **Lint and existing targets.** `shellcheck` clean. `actionlint`
  (`rhysd/actionlint:latest`) clean apart from the `linux-amd64-cpu4`
  unknown-runner-label warning every existing workflow here carries. `yamllint`
  clean. `make build`, `make test`, `make check-vendor` and
  `make -C deployments/devel check-modules` pass. The distroless image was built
  for `linux/arm64` to confirm nothing here affects it.


rahulait (approved):

.github/workflows/third-party-notices-check.yaml
```diff
@@ -0,0 +1,61 @@
...
+
+# Regenerates THIRD_PARTY_NOTICES.md and fails if it differs from the committed
+# copy, so a dependency change cannot land without refreshed attribution.
+#
+# This runs on every build rather than behind a changed-paths filter. The
+# runtime inventory is the import closure of ./cmd/..., so it changes when
+# ordinary .go files change their imports, not only when go.mod or vendor/ move.
+# A filter keyed on dependency manifests would therefore pass green while the
+# committed file went stale, and the failure would surface on main instead of on
+# the pull request that caused it.
```
tariq1890:
```suggestion
```

Makefile
```diff
@@ -53,6 +53,32 @@ goimports:
...
 
+# Regenerate THIRD_PARTY_NOTICES.md from vendor/ and deployments/devel/go.mod.
+#
+# go-licenses is pinned next to the rest of the build toolchain in
+# deployments/devel/go.mod. Install it into ./bin rather than relying on
+# whatever is on PATH, so a developer's globally installed copy cannot change
+# the generated file. It depends on the pin, or bumping the version would leave
+# a stale binary here while CI builds the new one from a fresh checkout, and the
+# only visible symptom would be a 'notices file is stale' failure nobody can
+# reproduce locally.
```
tariq1890: Let's remove

Makefile
```diff
@@ -53,6 +53,32 @@ goimports:
 lint:
 	golangci-lint run ./...
 
+# Regenerate THIRD_PARTY_NOTICES.md from vendor/ and deployments/devel/go.mod.
```
tariq1890: Can we make sure that the TPN doc doesn't include `/devel/go/mod`? This comment is suggesting otherwise

Makefile
tariq1890: let's not forget to remove this

tariq1890 (approved):

Makefile
abrarshivani: Fixed

Makefile
abrarshivani: Done

PR #240, opened 2026-08-20
Preserve pre-existing node cordons across driver upgrades

### LINKED ISSUE #212: [Bug]: Uncordon already cordoned nodes
**Describe the bug**
During the uninstall process, the driver manager cordons the node and it uncordons it at the very end of the process.
If the node is already cordoned at the beginning of this process, the driver manager should not uncordon it at the end. Cordons are used by Kubernetes administrators and Operators for various reasons and it should not be mandatory to remove the cordons at the end of this process (at least there should be a flag to opt out).

**To Reproduce**
- cordon a node
- launch the driver manager on it
- verify that the node in uncordoned

**Expected behavior**

 - `k8s-driver-manager` source (Commit SHA or image digest): main
 - Host OS: all
 - Kernel Version: all
 - NVIDIA Driver Version: all
 - Container Runtime Version: all
 - CPU Architecture: all
 - GPU Model(s): all

 If applicable, also provide:
 - Kubernetes Distro and Version: all
 - NVIDIA GPU Operator version: all

ben-feldstein:
Fixes #212.

## Problem

driver-manager uncordons the node unconditionally on its success path, in `cleanupOnFailure()`, and on the skip-eviction path that runs after a node reboot. It never checks whether the node was already cordoned before it acted. Any cordon placed by an administrator or another controller is therefore wiped whenever the driver pod is recreated, and on every node reboot. On the reboot path this is worst: driver-manager uncordons the node without ever having cordoned it.

The upgrade controller in [k8s-operator-libs](https://github.com/NVIDIA/k8s-operator-libs/blob/8e59492a772c39f8e99eea9cfc42567fcc1d939a/pkg/upgrade/common_manager.go#L264-L277) already solves this with a `node-initial-state.unschedulable` annotation and skips the uncordon for nodes that were cordoned before the upgrade started. driver-manager has no equivalent.

## Fix

I applied the same pattern to driver-manager.

- `CordonNode` first records the node's current schedulable state in a `nvidia.com/driver-manager.node-initial-state.unschedulable` annotation, then cordons. The annotation is written before the cordon and only when absent, so a restart of the init container after the cordon cannot overwrite the state recorded by the first run.
- `UncordonNode` restores the recorded state instead of blindly uncordoning: it uncordons only when the annotation records `false`, skips the uncordon when it records `true`, and removes the annotation in both cases. When the annotation is absent, driver-manager never cordoned the node in any run, so the node is left untouched.
- On an uncordon failure the annotation is kept, so a retry after the init container restarts can still uncordon.
- A recording found on a schedulable node is treated as stale and overwritten: a cycle interrupted after our cordon would have left the node unschedulable, so the annotation can only have survived a failed removal.
- The `shouldSkipUninstall` early return now also runs the state-aware uncordon, so a cordon left by a crashed prior cycle is released and the annotation cleaned up. With the annotation absent this is a no-op.

| Scenario | Before | After |
| --- | --- | --- |
| Node not cordoned, driver upgrade succeeds | uncordoned | uncordoned (unchanged) |
| Node not cordoned, drain fails | uncordoned, retry loop flaps the cordon | uncordoned, retry loop flaps the cordon (unchanged) |
| Node manually cordoned before upgrade | uncordoned | stays cordoned |
| Node manually cordoned, drain fails | uncordoned | stays cordoned |
| Node reboot with a manual cordon in place | uncordoned | stays cordoned |
| Init container restarts between cordon and uncordon | uncordoned | uncordoned, recorded state survives the restart |

## Notes

- One-time rollout note: a node that a previous driver-manager release cordoned mid-cycle carries no annotation, so this version will not auto-uncordon it. Operators should uncordon such nodes manually once after upgrading.
- A cordon placed by an external actor during the upgrade window is still restored to the recorded initial state at the end of the cycle. The same race exists in the upgrade controller.

## Testing

- Added unit tests for `CordonNode`/`UncordonNode` against a fake clientset covering: cordon/uncordon round trip on a schedulable node, a pre-cordoned node staying cordoned through the uncordon, the recorded state surviving an init container restart, a stale recording on a schedulable node being reconciled, an uncordon with no prior cordon leaving an external cordon in place, and the recording never being overwritten while the node is unschedulable.
- `go build ./...`, `go vet ./...` (linux), and `go test ./...` all pass.


internal/kubernetes/client.go
```diff
@@ -144,30 +125,102 @@ func (c *Client) GetNodeAnnotationValue(nodeName, annotation string) (string, er
...
+	}
+
+	patchBytes, err := json.Marshal(patch)
+	if err != nil {
+		return fmt.Errorf("failed to marshal patch: %w", err)
+	}
+
+	backoff := wait.Backoff{
+		Duration: time.Second,
+		Factor:   2.0,
```
kvalliyurnatt: nit: could you declare these values as constants

internal/kubernetes/client.go
```diff
@@ -144,30 +125,102 @@ func (c *Client) GetNodeAnnotationValue(nodeName, annotation string) (string, er
...
+		c.log.Infof("Node %s was already cordoned before the driver upgrade, skipping uncordon", nodeName)
+	} else {
+		c.log.Infof("Uncordoning node %s", nodeName)
+		drainHelper := &drain.Helper{Ctx: c.ctx, Client: c.clientset}
+		if err := drain.RunCordonOrUncordon(drainHelper, node, false); err != nil {
+			return err
+		}
+	}
+
+	return c.removeNodeAnnotation(nodeName, nodeInitialStateAnnotation)
```
kvalliyurnatt, line 180: If uncordoning succeeds but annotation removal fails, a stale "false" annotation remains. If an administrator later cordons the node, the next UncordonNode may interpret that stale annotation as ownership and remove the administrator’s cordon. Could the uncordon and annotation removal be performed atomically in one Kubernetes patch?

internal/kubernetes/client.go
```diff
@@ -144,30 +125,102 @@ func (c *Client) GetNodeAnnotationValue(nodeName, annotation string) (string, er
...
 	if err != nil {
 		return fmt.Errorf("failed to get node %s: %w", nodeName, err)
 	}
 
+	value, ok := node.Annotations[nodeInitialStateAnnotation]
+	initialState := strconv.FormatBool(node.Spec.Unschedulable)
+	if !ok || (!node.Spec.Unschedulable && value != initialState) {
+		c.log.Infof("Recording initial state of node %s in annotation %s=%s", nodeName, nodeInitialStateAnnotation, initialState)
+		if err := c.setNodeAnnotation(nodeName, nodeInitialStateAnnotation, initialState); err != nil {
+			return fmt.Errorf("failed to record initial state of node %s: %w", nodeName, err)
```
kvalliyurnatt: I think we should add a call to `cleanupOnFailure` at the caller https://github.com/NVIDIA/k8s-driver-manager/blob/f220c1350f07b5afacdcc7cb186247aca635efec/cmd/driver-manager/main.go#L418 when the cordon fails, since if the function fails on setting annotation it could leave components in a paused state, since the cordon itself would have succeeded at this point.

We could return a specific error to the caller to note annotation setting failed and call cleanup based on that.

kvalliyurnatt (changes requested):
Thanks for your contribution. Changes look mostly good, I think there is one thing that needs to be addressed

kvalliyurnatt:
This PR has been inactive for two weeks now, closing this in favor of https://github.com/NVIDIA/k8s-driver-manager/pull/258

PR #249, opened 2026-08-28
Manage the Go toolchain version with Renovate instead of Dependabot

kvalliyurnatt:
Manage the Go toolchain version with Renovate instead of Dependabot.

Dependabot’s Docker updater can select abbreviated tags such as 1.27, while the Go archive downloaded during the container build requires a full version such as 1.27.0. This caused the image build to request a nonexistent archive.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -1,0 +1,31 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "extends": [
+    "config:recommended",
+    ":disableDependencyDashboard"
+  ],
+  "forkProcessing": "enabled",
+  "ignorePaths": ["vendor/**"],
+  "enabledManagers": ["custom.regex"],
+  "recreateWhen": "always",
+  "pruneStaleBranches": true,
+  "rebaseWhen": "auto",
+  "customManagers": [
+    {
+      "customType": "regex",
+      "managerFilePatterns": [
+        "deployments/devel/Dockerfile"
+      ],
+      "matchStrings": [
+        "FROM golang:(?<currentValue>\\d+\\.\\d+\\.\\d+)"
+      ],
+      "depNameTemplate": "go",
+      "datasourceTemplate": "golang-version",
+      "versioningTemplate": "semver"
+    }
+  ],
+  "labels": [
+    "dependencies",
+    "renovate"
+  ]
+}
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -25,11 +25,6 @@ updates:
     schedule:
       interval: "daily"
 
-  - package-ecosystem: "docker"
-    directory: "/deployments/devel"
-    schedule:
-      interval: "daily"
-
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
--- a/.github/workflows/renovate.yaml
+++ b/.github/workflows/renovate.yaml
@@ -1,0 +1,22 @@
+name: Renovate
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 9 * * *" # every day at 9:00 AM UTC
+
+jobs:
+  renovate:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v7
+
+      - name: Run Renovate
+        uses: renovatebot/github-action@v46.2.2
+        with:
+          configurationFile: .github/renovate.json
+        env:
+          RENOVATE_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          RENOVATE_REPOSITORIES: '["NVIDIA/k8s-driver-manager"]'
+          RENOVATE_ONBOARDING: false


deployments/container/Dockerfile.distroless
```diff
@@ -12,6 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+ARG GOLANG_VERSION
+FROM golang:${GOLANG_VERSION} AS go-toolchain
```
cdesiniotis: Alternatively, could we just directly build in this image instead of the CUDA image? That would remove the need to copy over the toolchain. (and one less image for dependabot to raise version bump PRs for)

kvalliyurnatt: @cdesiniotis seems like we used to do that but ran into an issue https://github.com/NVIDIA/k8s-driver-manager/pull/12 that is when we changed to the current way of doing things.

henry118: I "think" we can simplify with following?

```
RUN dnf install -y make git gcc golang
RUN go env -w GOTOOLCHAIN="go${GOLANG_VERSION}"
```

kvalliyurnatt: @henry118 I think `GOTOOLCHAIN="go1.27"` does not work, it needs to be `GOTOOLCHAIN="go1.27.0"`. 

```
> go env -w GOTOOLCHAIN="go1.27"
> GOTOOLCHAIN=go1.27 go version
go: downloading go1.27 (darwin/arm64)
go: invalid toolchain: go1.27 is a language version but not a toolchain version (go1.27.x)
```

The current code I have allows for both `1.27` (which is what dependabot does) and `1.27.0`

rajathagasthya:
mig-parted has switched over to managing Go version using Renovate instead of Dependabot (https://github.com/NVIDIA/mig-parted/pull/467 describes why). We should consider doing the same here too, then we may not have to fix the Dockerfile stages. 

Also, with the current changes we're allowing Dependabot to pin an abbreviated tag such as 1.27. We should pin it to the full patch version instead so we know which Go version is being used.

kvalliyurnatt:
> mig-parted has switched over to managing Go version using Renovate instead of Dependabot ([NVIDIA/mig-parted#467](https://github.com/NVIDIA/mig-parted/pull/467) describes why). We should consider doing the same here too, then we may not have to fix the Dockerfile stages.
> 
> Also, with the current changes we're allowing Dependabot to pin an abbreviated tag such as 1.27. We should pin it to the full patch version instead so we know which Go version is being used.

Thanks @rajathagasthya I like that approach better, I have made changes to use renovate instead of dependabot.

.github/renovate.json
```diff
@@ -0,0 +1,31 @@
...
+  "pruneStaleBranches": true,
+  "rebaseWhen": "always",
+  "customManagers": [
+    {
+      "customType": "regex",
+      "managerFilePatterns": [
+        "deployments/devel/Dockerfile"
+      ],
+      "matchStrings": [
+        "FROM golang:(?<currentValue>\\d+\\.\\d+\\.\\d+)"
```
rajathagasthya, line 20: Let's add `ARG GOLANG_VERSION=1.26.6` in the Dockerfile so it's bumped accordingly.

.github/renovate.json
```diff
@@ -0,0 +1,31 @@
...
+  "ignorePaths": ["vendor/**"],
+  "enabledManagers": ["custom.regex"],
+  "recreateWhen": "always",
+  "pruneStaleBranches": true,
+  "rebaseWhen": "always",
+  "customManagers": [
+    {
+      "customType": "regex",
+      "managerFilePatterns": [
+        "deployments/devel/Dockerfile"
```
rajathagasthya, line 17: ~~This should be `deployments/container/Dockerfile.distroless`. Or both, if we care about this dev Dockerfile.~~

The `deployments/devel/Dockerfile` and `deployments/container/Dockerfile.distroless` setup is a bit weird. It looks like the former exists only to manage Go version.

deployments/container/Dockerfile.distroless
henry118: I see. Thanks for trying it out.

rajathagasthya:
@kvalliyurnatt Should we consider standardizing this configuration across all our repos? If so, let's create issues in the respective repos and add `help-wanted`/`good-first-issue` labels so we can encourage external contributions.

kvalliyurnatt:
@rajathagasthya I was going to ask the same question :) I think it makes sense, I will create the issues

.github/renovate.json
kvalliyurnatt: `deployments/devel/Dockerfile` already has the go version explicitly, the behavior in the  `deployments/container/Dockerfile.distroless` seems to be that a go version needs to be passed otherwise it fails, do we want to change that behavior ? the go version gets passed from `deployments/devel/Dockerfile`  to `deployments/container/Dockerfile.distroless` the way things are configured right now in our repo

.github/renovate.json
kvalliyurnatt: @rajathagasthya correct, I don't have context on why we do it this way, but that is why I thought we just have renovate update that one file and that should get passed down to the distroless docker file.

.github/renovate.json
rajathagasthya: I'm curious to know why this is necessary in this repo, but not in other repos. If there is an opportunity to simplify this and it's fairly straightforward, we should.

kvalliyurnatt: I agree, let me dig into this to figure out why this is the way it is.

kvalliyurnatt: @rajathagasthya seems like this pattern exists in a few of our other repos as well like the mig-parted repo you pointed at earlier
https://github.com/NVIDIA/mig-parted/blob/main/hack/golang-version.sh

Only difference there is in the container Dockerfile we add the Go version explicitly, but in CI that Dockerfile actually gets the Go version from the devel dockerfile similar to what happens here.

Similar thing exists in container toolkit as well.

From what I understand, we want one source of truth for the go version we use, so devel dockerfile seems to have been chosen for that and the version from that gets passed around

The only difference in the mig-parted repo, is we set the GO version in both dockerfiles.

rajathagasthya: Understood. Thanks for digging into this!

.github/renovate.json
```diff
@@ -0,0 +1,31 @@
...
+  "extends": [
+    "config:recommended",
+    ":disableDependencyDashboard"
+  ],
+  "forkProcessing": "enabled",
+  "ignorePaths": ["vendor/**"],
+  "enabledManagers": ["custom.regex"],
+  "recreateWhen": "always",
+  "pruneStaleBranches": true,
+  "rebaseWhen": "always",
```
rajathagasthya: non-blocking: I think we should reconsider rebasing always, which results in waste of CI time/resources. For example, I've had to `ok-to-test` a bunch of times in https://github.com/NVIDIA/gpu-driver-container/pull/976.

kvalliyurnatt: Makes sense, I changed this to on conflict so that it will not always rebase.

rajathagasthya: Should we just use `auto` (default) instead? https://docs.renovatebot.com/configuration-options/#rebasewhen

kvalliyurnatt: Done.

rajathagasthya (approved):
LGTM, thanks!

PR #254, opened 2026-09-02
vfio-manage: bind every PCI function of a GPU, not just the first auxiliary one

jjacobelli:
## Problem

On a host with a multi-function NVIDIA GPU, `vfio-manage bind` can leave the IOMMU group in a
state VFIO will not accept, and starting a VM then fails with:

```
vfio: group 15 is not viable
```

VFIO assigns an entire IOMMU group to a guest and refuses the group unless *every* device in it
is bound to a vfio driver or to no driver at all. A discrete NVIDIA GPU is a multi-function PCI
device:

| Function | Class    | Device                    |
|----------|----------|---------------------------|
| `.0`     | 0x030000 / 0x030200 | VGA / 3D controller |
| `.1`     | 0x040300 | HDA audio                 |
| `.2`     | 0x0c0330 | VirtualLink USB xHCI (Turing-era boards) |
| `.3`     | 0x0c8000 | USB-C UCSI (Turing-era boards) |

After `bind --all`, `.2` is typically still owned by `xhci_hcd`:

```
0000:16:00.0 -> vfio-pci
0000:16:00.1 -> vfio-pci
0000:16:00.2 -> xhci_hcd     <-- group is not viable
0000:16:00.3 -> UNBOUND
```

An *unbound* function is fine; a function owned by a host driver is not.

## Root cause

`getGraphicsAuxDev()` has three limitations that compound:

1. **It returns on the first match.** The `return auxDev, nil` is inside the loop, so at most one
   auxiliary function is ever bound.
2. **It discovers functions only via `consumer:pci:` device links.** The HDA audio function
   creates one; the VirtualLink xHCI and USB-C UCSI functions do not, so they are structurally
   invisible to this code no matter how many matches it collected.
3. **It returns early unless the device is `PCIVgaControllerClass`.** A GPU enumerating as a 3D
   controller (`0x030200`) gets no auxiliary handling at all.

## Fix

Replace it with `getAuxDevices()`, which returns every other PCI function belonging to the same
physical card. Functions are discovered two ways and the results unioned and de-duplicated:

- **sibling functions** sharing a `domain:bus:device` — the only way to reach the VirtualLink
  xHCI and USB-C UCSI functions;
- **`consumer:pci:` device links** — preserving the existing discovery path.

`BindToVFIODriver()` and `UnbindFromDriver()` now iterate over all of them.

Two deliberate constraints:

- **Siblings are scoped to the physical card, not to the IOMMU group.** Where ACS is unavailable
  a group can span an entire root port, and unrelated devices must not be unbound from their
  drivers.
- **A vendor check** guards against acting on anything that is not an NVIDIA function.

## Behaviour preserved

- Devices are still gated on `IsGPU()`, so **NVSwitches are unchanged** — they get no auxiliary
  handling, exactly as before. `IsGPU()` covers both the VGA and 3D controller classes, which is
  what fixes (3) above.
- A GPU whose only auxiliary function is HDA audio behaves exactly as it did.
- Consumer links are still followed wherever they point, so discovery is never narrower than
  before.
- A single-function GPU returns an empty set and takes the same path as today.
- No change to `cmd/vfio-manage/` — CLI behaviour and exit codes are untouched.

## Testing

Nine table-driven cases added in `internal/nvpassthrough/nvpassthrough_test.go`, using a fake
sysfs tree built with `t.TempDir()`:

| Case | Covers |
|---|---|
| Turing board: audio + VirtualLink xHCI + USB-C UCSI all returned | the bug |
| GPU with a single HDA audio function is still returned | pre-existing behaviour |
| Auxiliary device reachable only via a consumer link is still found | discovery not narrowed |
| 3D-controller-class GPU has its auxiliary functions returned | defect (3) |
| NVSwitch has no auxiliary functions handled | behaviour preserved |
| Single-function GPU has no auxiliary devices | no spurious results |
| Sibling function from another vendor is not touched | vendor guard |
| Functions on a different slot are not siblings | card-scoped, not group-scoped |
| A function found both as a sibling and via a consumer link appears once | de-duplication |

Plus `TestGetAuxDevicesReportsCurrentDriver`, since `BindToVFIODriver` skips functions already on
`vfio-pci` and unbinds the rest.

Verified locally:

```
go build ./...   OK
go vet ./...     clean
gofmt            clean
go test ./...    ok
golangci-lint    0 issues
```

---

*This change was written with AI assistance (Claude) and reviewed by me.*

### DIFF
--- a/internal/nvpassthrough/nvpassthrough.go
+++ b/internal/nvpassthrough/nvpassthrough.go
@@ -34,6 +34,7 @@ const (
 	pciDriversRoot    = pciRootDir + "drivers"
 	vfioPCIDriverName = "vfio-pci"
 	consumerPrefix    = "consumer:pci:"
+	nvidiaVendorID    = "0x10de"
 	libModulesRoot    = "/lib/modules/"
 )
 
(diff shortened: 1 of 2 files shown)


cdesiniotis (approved):
LGTM. Thank you @jjacobelli for the contribution!

If you are feeling up to it, we could also use this change in https://github.com/NVIDIA/go-nvlib/tree/main/pkg/nvpassthrough. We have been meaning to update k8s-driver-manager to import this package now from go-nvlib (the `dra-driver-nvidia-gpu` uses this code as well when setting up GPUs for passthrough).

PR #256, opened 2026-09-02
Add go modernize check and fix any existing issues

kvalliyurnatt:
Adding the modernize lint rule so that any future changes that do not match the newer go standards/patterns are caught, also fixing a couple of existing things by running go fix


deployments/devel/Makefile
```diff
@@ -19,6 +19,7 @@ download:
 install-tools: download
 	@echo Installing tools from tools.go
 	@cat tools.go | grep _ | awk -F'"' '{print $$2}' | xargs -tI % go install %
+	@go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint
```
tariq1890: Can we pin this to a particular version?

kvalliyurnatt: Pinned the install command to golangci-lint v2.13.2

deployments/devel/Makefile
```diff
@@ -19,6 +19,7 @@ download:
 install-tools: download
 	@echo Installing tools from tools.go
 	@cat tools.go | grep _ | awk -F'"' '{print $$2}' | xargs -tI % go install %
+	@go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.13.2
```
tariq1890: Can we move the golangci-lint installation to the root Makefile instead? The version pinning of golangci-lint can be done through `versions.mk`

kvalliyurnatt: Moved golangci-lint installation to the root Makefile and pinned v2.13.2 through versions.mk. I also removed it from the development tooling module.

Makefile
```diff
@@ -50,12 +50,17 @@ goimports:
...
 		| xargs goimports -local $(MODULE) -w
 
-lint:
-	golangci-lint run ./...
-
 BIN_DIR := $(CURDIR)/bin
+GOLANGCI_LINT := $(BIN_DIR)/golangci-lint
 GO_LICENSES := $(BIN_DIR)/go-licenses
 
+$(GOLANGCI_LINT): versions.mk
```
tariq1890: Is this invocation of `versions.mk` needed here? We already include it (see L21).
```suggestion
$(GOLANGCI_LINT):
```

.golangci.yml
```diff
@@ -0,0 +1,5 @@
+version: "2"
+
+linters:
+  enable:
+    - modernize
```
rajathagasthya, line 9: Can we match linters from other repos? Okay to do in a follow up.

tariq1890: agreed, I was going to post the same comment. Doing it in a follow-up sounds good to me.

Makefile
kvalliyurnatt: The include on L21 loads the variables, while this prerequisite invalidates the generated bin/golangci-lint target when versions.mk changes. Without it, updating GOLANGCI_LINT_VERSION could leave an existing stale binary in place. This might not be an issue in CI I guess, only when run locally. I am ok to remove it, don't feel strongly about it.

.golangci.yml
kvalliyurnatt: I went ahead and included this in the existing PR

Makefile
tariq1890: In that case, maybe we are better off just declaring the golangci-lint version in the root `Makefile`?

kvalliyurnatt: done.

rajathagasthya (approved):

tariq1890 (approved):
Thanks @kvalliyurnatt !

Makefile
```diff
@@ -50,12 +50,18 @@ goimports:
 	go list -f {{.Dir}} $(MODULE)/... \
 		| xargs goimports -local $(MODULE) -w
 
-lint:
-	golangci-lint run ./...
-
 BIN_DIR := $(CURDIR)/bin
+GOLANGCI_LINT_VERSION ?= v2.13.2
```
abrarshivani, line 54: We pin golangci-lint to v2.13.2 for make lint, but the golang ci workflow still uses golangci-lint-action with version: latest. Can we pin the ci action to the same version as well, or have ci use make lint, so local and ci lint results stay consistent?

abrarshivani (approved):
LGTM. Just one comment about keeping the golangci-lint version consistent with CI.

Makefile
kvalliyurnatt: Fixed this, thanks!

rahulait: Can we keep the GOLANGCI_LINT_VERSION in one file (versions.mk) and read from there in Makefile and github workflow?

kvalliyurnatt: @rahulait  sorry i missed this and merged, I can open a separate PR to address this

rahulait: Thats fine, we can address this later on :)

PR #258, opened 2026-09-08
Add annotations to record if node was cordoned by the driver manager

kvalliyurnatt:
Currently the driver manager uncordons the node unconditionally, so it can end up uncordoning nodes that were not cordoned by the driver manager. We already address a similar scenario in the upgrade manager on the GPU operator by adding an annotation on the nodes to denote the initial state of a node before the upgrade manager cordons the node, this PR implements a similar solution to work around this issue. This is the first part of the solution where we add an annotation to denote if the node was cordoned by the k8s-driver-manager, we will enforce the cordoning behavior in a subsequent release.

uses fixes proposed as part of https://github.com/NVIDIA/k8s-driver-manager/pull/240 .

part of addressing: #212


internal/kubernetes/client_test.go
```diff
@@ -0,0 +1,254 @@
...
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"testing"
+
+	"github.com/sirupsen/logrus"
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
```
ben-feldstein, line 27: Needed for the test suggested at the bottom of this file.

```suggestion
	corev1 "k8s.io/api/core/v1"
	apierrors "k8s.io/apimachinery/pkg/api/errors"
```

ben-feldstein (commented):
The single patch is better than what I had on #240. Pinning `resourceVersion` and re-reading the node on conflict closes the window.

A few comments below. The missing conflict test is the only one I'd call important.

One other thing: could you add a co-author trailer to the commit? The approach, the `main.go` change, and the test scaffolding came from #240, and `newTestClient`, `getTestNode`, `TestCordonUncordonNode` and `TestCordonNodeRecordsInitialState` are unchanged from that branch.

```
Co-authored-by: Ben Feldstein <ben.feldstein@baseten.co>
```

Either way, glad it's landing.

internal/kubernetes/client_test.go
```diff
@@ -0,0 +1,254 @@
...
+
+func TestUncordonNodeRejectsInvalidRecordedState(t *testing.T) {
+	client, clientset := newTestClient(newTestNode(
+		true,
+		map[string]string{nodeInitialUnschedulableAnnotationKey: "invalid"},
+	))
+
+	require.Error(t, client.UncordonNode(testNodeName))
+	require.Empty(t, getNodePatches(t, clientset))
+}
```
ben-feldstein, line 322: No test returns a 409, so the `IsConflict` retry never runs. The fake tracker doesn't check `resourceVersion` on update either, so `TestNodeSchedulingStatePatchesAreAtomic` proves the patch body is right but not that a conflicting write gets retried.

This covers both directions. I ran it against this branch: gofmt and vet clean, passes, and it fails if `IsConflict` is replaced with a predicate that never retries.

```suggestion
}

func TestCordonUncordonNodeRetriesOnConflict(t *testing.T) {
	testCases := []struct {
		name   string
		cordon bool
	}{
		{name: "cordon", cordon: true},
		{name: "uncordon"},
	}

	for _, tc := range testCases {
		t.Run(tc.name, func(t *testing.T) {
			node := newTestNode(!tc.cordon, nil)
			if !tc.cordon {
				node.Annotations = map[string]string{nodeInitialUnschedulableAnnotationKey: "false"}
			}
			client, clientset := newTestClient(node)

			conflicts := 0
			clientset.PrependReactor("patch", "nodes", func(k8stesting.Action) (bool, runtime.Object, error) {
				if conflicts == 0 {
					conflicts++
					return true, nil, apierrors.NewConflict(
						corev1.Resource("nodes"), testNodeName, errors.New("object was modified"),
					)
				}
				return false, nil, nil
			})

			if tc.cordon {
				require.NoError(t, client.CordonNode(testNodeName))
			} else {
				require.NoError(t, client.UncordonNode(testNodeName))
			}

			require.Equal(t, 1, conflicts)
			require.Len(t, getNodePatches(t, clientset), 2)
			require.Equal(t, tc.cordon, getTestNode(t, clientset).Spec.Unschedulable)
		})
	}
}
```

internal/kubernetes/client.go
```diff
@@ -144,30 +147,91 @@ func (c *Client) GetNodeAnnotationValue(nodeName, annotation string) (string, er
...
+}
+
+func nodeUpdateBackoff() wait.Backoff {
+	return wait.Backoff{
+		Duration: time.Second,
+		Factor:   2.0,
+		Jitter:   0.2,
+		Steps:    7,
+	}
 }
```
ben-feldstein: `UpdateNodeLabels` still has its own inline `wait.Backoff` with these exact values, so there are two copies now. Worth pointing it at this helper. That would also force a decision on the predicate: `UpdateNodeLabels` retries on any error, this path only on conflict.

Minor: `Duration` is a full second, so a test that forces a conflict really sleeps. The one I suggested takes about 2.3s for two subtests. A package-level `var` instead of a func would let tests shorten it.

internal/kubernetes/client.go
```diff
@@ -144,30 +147,91 @@ func (c *Client) GetNodeAnnotationValue(nodeName, annotation string) (string, er
...
 
-	node, err := c.clientset.CoreV1().Nodes().Get(c.ctx, nodeName, metav1.GetOptions{})
-	if err != nil {
-		return fmt.Errorf("failed to get node %s: %w", nodeName, err)
-	}
+		recordedState, recorded := node.Annotations[nodeInitialUnschedulableAnnotationKey]
+		if node.Spec.Unschedulable && recorded {
+			if _, err := strconv.ParseBool(recordedState); err != nil {
+				return fmt.Errorf("invalid value %q for node annotation %s: %w", recordedState, nodeInitialUnschedulableAnnotationKey, err)
+			}
```
ben-feldstein, line 159: Is the hard failure here intentional? `TestUncordonNodeRejectsInvalidRecordedState` covers the uncordon side, but this path is different: `IsConflict` won't retry a parse error and `cmd/driver-manager/main.go:418` returns it, so a bad value here fails the driver upgrade on that node until someone fixes the annotation by hand.

Only driver-manager writes this key, so getting here means it was hand-edited, and failing closed on that is reasonable. The alternative is to treat the value as garbage: leave `spec.unschedulable` alone and clear the annotation. Either is fine, I just want to know it was chosen.

internal/kubernetes/client_test.go
kvalliyurnatt: Added conflict retry coverage for both CordonNode and UncordonNode. The test forces one 409 and verifies the second patch succeeds.

internal/kubernetes/client_test.go
kvalliyurnatt: Added the apierrors import for the conflict retry test.

internal/kubernetes/client.go
kvalliyurnatt: Yes, failing closed is intentional: an invalid ownership record should not be overwritten or used to change scheduling state automatically. I expanded the test to cover both CordonNode and UncordonNode and verify that neither patches the node.

internal/kubernetes/client.go
kvalliyurnatt: Consolidated the duplicate settings into a package-level nodeUpdateBackoff variable, and the conflict test temporarily sets its duration to zero. UpdateNodeLabels intentionally retains its existing retry-on-any-error behavior to avoid changing its semantics in this fix.

kvalliyurnatt:
Added the requested co-author trailer  to the commit.

ben-feldstein (approved):

PR #263, opened 2026-09-10
Invalidate status files before rescheduling operands

rajathagasthya:
The operator-validator and driver containers remove the readiness files under `/run/nvidia/validations` in their preStop hooks. Those hooks do not run when the containers are removed outside the kubelet, for example when a CRI-O minor version upgrade wipes every pod sandbox on the node. The files survive on the host tmpfs. When k8s-driver-manager then re-enables the operand labels, the operand pods are recreated together with the driver pod, before the driver is installed. A stale `toolkit-ready` lets an operand init container pass while the kernel module is unloaded, so the nvidia runtime creates the main container without the driver libraries. The MIG manager then fails NVML initialization with `ERROR_LIBRARY_NOT_FOUND` and leaves the node at `mig.config.state=failed` until the pod is deleted.

Remove the files once the previous driver has been torn down, just before the operands are rescheduled: after the stale rootfs is unmounted on the fast path, and after the kernel modules are unloaded and the rootfs unmounted on the full path. The validators were deleted by the eviction before that, so nothing recreates the files in between. Every failure that can still fall back to the loaded driver, such as a drain failure or modules in use, happens before the removal, so `cleanupOnFailure` re-enables the operands with the files it finds, as it does today. The paths that do not evict, because no driver is loaded or because eviction itself failed, leave the files alone: a validator that is still running would not rewrite them.

The set removed is what the two preStop hooks remove: the `*-ready` files written by the validators, plus the driver's `.driver-ctr-ready` and `.driver-daemons-status`. Other hidden markers, such as cc-manager's `.cc-manager-ctr-ready`, are left in place because k8s-driver-manager does not restart their owners. The sandbox validator waits on that marker before writing `cc-manager-ready`, and cc-manager can share a node with the vfio-manager DaemonSet, which also runs this init container. `workload-type` is not a readiness marker and is also left in place.

In a normal driver rollout both preStop hooks have already run and the removal is a no-op. On the fast path that skips the module unload, removing `.driver-ctr-ready` means the validator waits for the new driver container's startup probe to rewrite it, which is the ordering a normal rollout already has.

Removal failures are logged and ignored in order to not abort driver rollout.

Related: NVIDIA/gpu-operator#2874. The gpu-operator side, gating the operand init containers on the kernel module, is NVIDIA/gpu-operator#2881.


cmd/driver-manager/main.go
```diff
@@ -369,6 +370,8 @@ func (dm *DriverManager) uninstallDriver() error {
 		return fmt.Errorf("failed to evict GPU operator components: %w", err)
 	}
 
+	dm.invalidateStatusFiles(validationsDir)
```
cdesiniotis, line 373: Question -- doesn't the race still exist if we call this method after evicting the GPU Operator components? I assume there is a still a small window where an operand pod can start before the status file is deleted. Should we call this method before `evictAllGPUOperatorComponents`?

cmd/driver-manager/main.go
```diff
@@ -902,6 +905,34 @@ func (dm *DriverManager) removePIDFile() {
 	}
 }
 
+// invalidateStatusFiles removes the readiness files that the validator and
+// driver preStop hooks remove, which does not happen when those containers are
+// removed outside the kubelet. Hidden markers of components that
+// k8s-driver-manager does not restart, such as .cc-manager-ctr-ready, are kept.
+func (dm *DriverManager) invalidateStatusFiles(dir string) {
```
cdesiniotis, line 912: As an alternative, should we consider the below? 
```
os.RemoveAll("/run/nvidia/validations")
os.Mkdir("/run/nvidia/validations", 0755)
```

cmd/driver-manager/main.go
rajathagasthya: I don't think the race exists with this code. Eviction sets the operand labels to `paused-for-driver-upgrade` which terminates the pods. They are not recreated until `rescheduleGPUOperatorComponents` sets the labels back to `true`. The files are removed after eviction and before the operands are rescheduled.

In fact, calling this method before eviction is risky. If eviction fails, `cleanupOnFailure` re-enables the labels and a running validator pod (failed eviction) keeps running without re-creating the status files. So any operand pods created after that would be stuck at `Init:0/1` until validator is restarted and new status files are written.

cmd/driver-manager/main.go
rajathagasthya: I considered that, but it would also delete `.cc-manager-ctr-ready` file. As far as I know, nothing re-creates that file unlike the other status files, and `cc-manager` is also not in the driver-manager's eviction list. So I'm not entirely sure if it's safe to delete.

cdesiniotis: Ah okay, gotcha. Let's stick with the current implementation then.

cmd/driver-manager/main.go
cdesiniotis: My fault, I mixed up `eviction` with `rescheduling` of the operands when reviewing this. Ignore my initial comment.

tariq1890: Do we want to call this method during the fast path though? 

There won't be any resetting of operands when the fast path is triggered, right?

rajathagasthya: I see fast path also evicts and reschedules the operands. I think only the driver installation is skipped when kernel modules are loaded. I confirmed this on a cluster too. https://github.com/NVIDIA/k8s-driver-manager/blob/63085896cbf9927c6dcb5971fb143a67759e3c1e/cmd/driver-manager/main.go#L362-L401

We could skip deleting status files during fast-path restart, but the operands are evicted, so not having respective status files would be an accurate reflection of their states.

tariq1890: Ahh, got it. Thanks for pointing that out.

> so not having respective status files would be an accurate reflection of their states.

Agreed. I will resolve this comment

tariq1890: Per Devin AI, the deletion of the status files may end up being premature in case of any error thrown in the control flow between this method and the end of the `uninstallDriver` method. Right now any failure in the  uninstallMethod is followed by a method call to `cleanupOnFailure` which in turn calls the `rescheduleGPUOperatorComponents` method. Since `rescheduleGPUOperatorComponents` results in the operands coming back up, that may now fail since the status files no longer exist owing to the earlier `invalidateStatusFiles` call.

rajathagasthya: I think this is a valid concern. If `.driver-ctr-ready` is gone and there are any errors between this method and `uninstallDriver`, the validator's `nvidia-validator COMPONENT=driver` init container can be stuck waiting for this file. This file is only written by the driver container's startup probe, which does not start until this init container succeeds. So this edge case is possible. 

Let me see how I can address this. Thanks for flagging!

